PR #3, opened 2016-09-21
Explaining weights for linear regression models

lopuhin:
The "classifier" key is still used in `explain_weights` result even for regression models. Not sure if we need a better name - maybe "model" is more general?

### DIFF
--- a/eli5/sklearn/explain_weights.py
+++ b/eli5/sklearn/explain_weights.py
@@ -3,9 +3,9 @@ from __future__ import absolute_import
 from singledispatch import singledispatch
 
 import numpy as np
-from sklearn.linear_model import LogisticRegression, LogisticRegressionCV
-from sklearn.linear_model import SGDClassifier, Perceptron
-from sklearn.linear_model import PassiveAggressiveClassifier
+from sklearn.linear_model import (
+    LogisticRegression, LogisticRegressionCV, PassiveAggressiveClassifier,
+    Perceptron, SGDClassifier,  SGDRegressor)
 from sklearn.svm import LinearSVC
 # TODO: see https://github.com/scikit-learn/scikit-learn/pull/2250
 from sklearn.naive_bayes import BernoulliNB, MultinomialNB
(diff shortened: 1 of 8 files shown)


kmike:
scikit-learn uses 'estimator' name, what do you think about switching to it?

lopuhin:
Yeah, I like it!

eli5/sklearn/explain_weights.py
```diff
@@ -219,6 +235,90 @@ def explain_tree_feature_importance(clf, vec=None, top=_TOP, class_names=None,
...
+    Return an explanation of a linear regressor weights in the following
+    format::
+
+        {
+            "estimator": "<regressor repr>",
+            "method": "<interpretation method>",
+            "description": "<human readable description>",
+            "targets": [
+                {
+                    "targets <target name>,
```
kmike: formatting looks off

eli5/sklearn/utils.py
```diff
@@ -29,28 +37,49 @@ def get_feature_names(clf, vec=None, bias_name='<BIAS>', feature_names=None):
...
+
+
 def get_coef(clf, label_id):
     """
     Return a vector of coefficients for a given label,
     including bias feature.
     """
-    coef = clf.coef_[label_id]  # multiclass case, also works for binary
+    if len(clf.coef_.shape) == 2:
+        coef = clf.coef_[label_id]  # multiclass case, also works for binary
```
kmike: it seems the comment is incorrect if you needed this change

lopuhin: Right, SGDRegressor stores coefficients as a one-dim array, and binary does not make sense here, I'll fix comments

kmike: it seems it only worked for sparse `coef_` matrices, and tests used only such matrices

kmike:
Looks good! Could you please also remove a TODO item in README?

eli5/sklearn/utils.py
lopuhin: I think most (or all) classifiers we test have dense ``coef_`` matrices?

eli5/sklearn/explain_weights.py
lopuhin: Thanks for the catch! Fixed

lopuhin:
I updated TODO, also mentioned that explaining predictions is currently not supported for regression models and trees (working on regression models at the moment).

kmike:
Thanks, looks great!

PR #4, opened 2016-09-21
Vec not required for explain prediction

lopuhin:
Not that I hate vectorizers :)
But ``vec`` is not strictly required for ``explain_prediction`` (in the same way it is not required for explain_weights), so I think it makes sense to have ``vec`` after ``doc``. I'm not 100% sure about 8bcb706 though - with it we need to pass vectorized=True only if we pass vec and already vectorized doc, but don't have to do it if we don't pass vec - not sure if it's too magical.

### DIFF
--- a/eli5/sklearn/explain_prediction.py
+++ b/eli5/sklearn/explain_prediction.py
@@ -23,8 +23,8 @@ _TOP = 20
 
 
 @singledispatch
-def explain_prediction(clf, doc, vec=None, top=_TOP, class_names=None,
-                       feature_names=None, target_names=None):
+def explain_prediction(clf, doc, vec=None, top=_TOP, target_names=None,
+                       feature_names=None):
     """ Return an explanation of an estimator """
     return {
         "estimator": repr(clf),
@@ -39,7 +39,7 @@ def explain_prediction(clf, doc, vec=None, top=_TOP, class_names=None,
 @explain_prediction.register(Perceptron)
 @explain_prediction.register(LinearSVC)
 def explain_prediction_linear(
-        clf, doc, vec=None, top=_TOP, class_names=None,
+        clf, doc, vec=None, top=_TOP, target_names=None,
         feature_names=None, vectorized=False):
     """ Explain prediction of a linear classifier. """
     feature_names = get_feature_names(clf, vec, feature_names=feature_names)
@@ -67,7 +67,7 @@ def explain_prediction_linear(
         return get_top_features_dict(feature_names, scores, top)
 
     def _label(label_id, label):
-        return rename_label(label_id, label, class_names)
+        return rename_label(label_id, label, target_names)
 
     if is_multiclass_classifier(clf):
         for label_id, label in enumerate(clf.classes_):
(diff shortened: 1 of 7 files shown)


eli5/sklearn/utils.py
```diff
@@ -35,7 +35,7 @@ def get_feature_names(clf, vec=None, bias_name='<BIAS>', feature_names=None):
     """
     if feature_names is None:
         if vec and hasattr(vec, 'get_feature_names'):
-            feature_names = vec.get_feature_names()
+            feature_names = list(vec.get_feature_names())
```
lopuhin: Forgot to add - this change is required because for some some vectorizers (DictVectorizer) get_features_names returns a list that holds the feature names for the vectorizer, and we append to it bias several lines below and modify it.

eli5/sklearn/explain_prediction.py
```diff
@@ -111,3 +106,13 @@ def _add_intercept(X):
         return sp.hstack([X, intercept]).tocsr()
     else:
         return np.hstack([X, intercept])
+
+
+def _get_X(doc, vec=None, vectorized=False):
```
kmike: heh, I extracted exactly the same utility (not commited) and named it almost the same (`_get_X_dense`)

lopuhin: :) I hope I didn't change too many unrelated things! I don't have any more global changes in the pipeline, just adding prediction explanation for regressors.

eli5/sklearn/explain_prediction.py
```diff
@@ -23,8 +23,8 @@
 
 
 @singledispatch
-def explain_prediction(clf, vec, doc, top=_TOP, class_names=None,
-                       feature_names=None, vectorized=False):
+def explain_prediction(clf, doc, vec=None, top=_TOP, class_names=None,
+                       feature_names=None, target_names=None):
```
kmike: probably it makes sense to rename class_names to target_names and keep a single argument for them; target_names is the name used in sklearn.

kmike:
The change makes sense; I experimented with something similar, but stopped because wanted clf, vec arguments to be together, and doc can't be a first argument because of singledispatch. But it is a stupid reason to require vec argument.

eli5/sklearn/explain_prediction.py
lopuhin: Yeah, that would be better, I like it!

lopuhin:
Yep, clf and vec look nicer together. It shouldn't be too hard to have doc the first argument, by just introducing an intermediate function and doing singledispatch through it - what do you think?

kmike:
> Yep, clf and vec look nicer together. It shouldn't be too hard to have doc the first argument, by just introducing an intermediate function and doing singledispatch through it - what do you think?

I think it doesn't worth it; the API in this PR looks fine.

lopuhin:
Ok, I'll just change the class_names to target_names then.

eli5/sklearn/explain_prediction.py
```diff
@@ -23,8 +23,8 @@
 
 
 @singledispatch
-def explain_prediction(clf, vec, doc, top=_TOP, class_names=None,
-                       feature_names=None, vectorized=False):
+def explain_prediction(clf, doc, vec=None, top=_TOP, target_names=None,
+                       feature_names=None):
```
kmike: why is "vectorized" argument removed?

lopuhin: oh, sorry - I removed it by accident, fixed now. Thanks for spotting that!

lopuhin:
Thanks! :)

PR #6, opened 2016-09-22
removed inherited configuration

stevepeak:
The actual issue was only the `bot: null`. Sorry about that.

fix https://github.com/codecov/support/issues/296


kmike:
That's awesome, thanks a lot!

PR #7, opened 2016-09-23
[WIP] InverseFeatureHasher and InverseHashingVectorizer

kmike:
TODO:

* [x] integrational tests
* [ ] unit tests  - moved to https://github.com/TeamHG-Memex/eli5/issues/8
* [x] invert coefficients for features which are negative
* [x] don't show signs for features if a coefficient is inverted
* [x] ~~expose features to formatters, don't just join feature names?~~
* [x] show hashing-specific caveats somehow
* [ ] check it with explain_prediction (.fit can be automated) - moved to https://github.com/TeamHG-Memex/eli5/issues/9
* [ ] in case of collisions sort features by their counts in training dataset? - moved to https://github.com/TeamHG-Memex/eli5/issues/10
* [x] partial_fit method
* [x] are zero features really filtered out?

Usage example:

```py
from sklearn.datasets import fetch_20newsgroups
from sklearn.feature_extraction.text import HashingVectorizer
from sklearn.linear_model import LogisticRegression

from eli5.sklearn import explain_prediction, explain_weights, InvertableHashingVectorizer
from eli5.formatters import format_as_text

categories = ['alt.atheism', 'comp.graphics']
remove = ('headers', 'footers', 'quotes')
data_train = fetch_20newsgroups(subset='train', categories=categories,
                                shuffle=True, random_state=42,
                                remove=remove)
X_raw = data_train.data
y = data_train.target

# use smaller n_features to get collisions
vec = HashingVectorizer(ngram_range=(1, 2), n_features=100000)
clf = LogisticRegression()
X = vec.fit_transform(X_raw)
clf.fit(X, y)

ivec = InvertableHashingVectorizer(vec)
ivec.fit(X_raw[:100])  # all documents are better, but they are not required

print(format_as_text(explain_weights(clf, ivec)))
```

Currently output looks like this:

```
Explained as: linear model

Features with largest coefficients.
Caveats:
1. Be careful with features which are not
   independent - weights don't show their importance.
2. If scale of input features is different then scale of coefficients
   will also be different, making direct comparison between coefficient values
   incorrect.
3. Depending on regularization, rare features sometimes may have high
   coefficients; this doesn't mean they contribute much to the
   classification result for most examples.

Feature names are restored from their hashes; this is not 100% precise
because collisions are possible. For known collisions possible feature names
are separated by | sign. Keep in mind the collision list is not exhaustive.
Features marked with (-) should be read as inverted: if they have positive
coefficient, the result is negative, if they have negative coefficient,
the result is positive.

y=1 top features
-----------------------------------
         (-)that | by any   +3.099
                 graphics   +2.515
                      for   +2.027
approximately | (-)people   +1.784
                    image   +1.547
                   thanks   +1.475
                     with   +1.410
                       on   +1.397
                    files   +1.389
                     file   +1.383
                  program   +1.327
                      any   +1.294
                      ...   (35136 more positive features)
                      ...   (34007 more negative features)
                       no   -1.305
                     they   -1.528
                      who   -1.550
        (-)criticize | he   -1.558
                       of   -1.597
                      god   -1.788
                      not   -1.981
              you | great   -2.291
```


lopuhin:
Maybe it would be helpful to have a partial_fit for the inverted vectorized? Since hashing vectorizer is often used in on-line fashion.

kmike:
A good idea! 

Initially I thought about a different use pattern: you have a large dataset, and you want to learn on it online; this makes CountVectorizer unfeasible. So you use HashingVectorizer. Then you get a sample of documents (not all) and use them to fit inverse hasher.

But now I see how partial_fit can be helpful; it can be also implemented more efficiently than CountVectorizer - it can just add seen terms to a set (or maybe a trie?), and compute term signs, indices, collisions etc. only on demand.

PR #21, opened 2016-10-13
Add format_as_html

lopuhin:
This PR adds ``eli5.formatters.format_as_html``. ``format_as_text`` is moved into ``eli5.formatters.text``, but is accessible by the old name ``eli5.formatters.format_as_text``. HTML formatting is done with Jinja2. This seems to be the most popular standalone templating library. But maybe just generating xml with etree from the standard library would be better (no need to learn another language) - what do you think @kmike ?

There is one thing that I punted on here: caveats and detailed explanations are rendered in a ``<pre>`` block. Probably a proper solution would be to represent them in some structured way, and then format for terminal or html display separately. I didn't do it because rendering in ``<pre>`` is easy and looks not too bad IMO.

html created during test runs are saved to ``.html`` folder, with names that include test module and the caller.


lopuhin:
I still want to simplify tests a bit and check trees explanations

kmike:
Hm, I was thinking about sending JSON with explanations and rendering them on a client. But using Jinja2 looks more lightweight; the dependency sounds fine.

Some features which can be nice to have:

* 'show more features' button;
* scroll long feature names (e.g. HashingVectorizer with many collisions looks bad in terminal);
* color bars which represent weights; longer bar -> larger absolute weight;
* show/hide description and caveats, hide them by default, to make display more compact. Maybe a '?' button with explanation popup.
* big one: for prediction explanations highlight something in input data, depending on input data type - e.g. words in text.

Do you think Jinja2 + JS snippets will scale well to these tasks, or what will be easierto develop?

lopuhin:
Yeah, I did not consider making html display more usable and convenient - it might be better to have a different implementation. I think it makes sense to implement some of these features and see which method is more convenient.

I think it's valuable for html output to be standalone (just an html snippet that can be included directly multiple times), but rendering with JS does not conflict with it, because we can embed JSON data and use CDNs for JS libraries. Having an ability to display up to 10K features also looks doable with this approach (that could mean about 100-500KB of JSON, depending on the features). But supporting larger number of features or more interactivity then just filtering might require a client-server architecture.

kmike:
I've tried the current HTML output; it works nice in standalone files, but in IPython plain-text version looks better to me:



vs 



Plaintext version uses 2x less vertical space, weights are close to terms, there is consistent font and there are no styling issues. Currently advantage of HTML version is that if feature names are large (e.g. when there is a lot of collisions) then layout doesn't break, and that it can be styled with CSS in dashboards.

lopuhin:
Yeah, I also like text version better, HTML is definitely not much usable yet :)
I'm working on a prototype of the feature highlighting in text, and will also tweaks styles, both standalone and in IPython.

Btw, speaking of weights, what do you think about swapping weights and features columns, so that the most important feature is closes to the weight?

kmike:
> Yeah, I also like text version better, HTML is definitely not much usable yet :)
I'm working on a prototype of the feature highlighting in text, and will also tweaks styles, both standalone and in IPython.

Great, mapping from features to text for sklearn vectorizers should be a cool feature! 
As for styles, what do you think about making separate functions for rendering HTML and for returning styles? If user wants to embed HTML snippets styles should be embedded only once; also, sometimes user may want to use custom styles.

> Btw, speaking of weights, what do you think about swapping weights and features columns, so that the most important feature is closes to the weight?

A good idea! It looks even more useful for plain text formatter. For HTML we should add better indication of feature weights anyways (bar plot?).

lopuhin:
> As for styles, what do you think about making separate functions for rendering HTML and for returning styles? If user wants to embed HTML snippets styles should be embedded only once; also, sometimes user may want to use custom styles.

Yeah, it's even much more convenient for developing, as there is just one set of styles to update.
Github also strips all styles when displaying notebooks (that certainly makes sense), maybe an option with inline styles would also be good to have.

kmike:
Text highlighting looks great!

What do you think about adding an option to use opacity in addition to color for highlighting? 

```py
def _highlight(doc, weighted_spans):
    char_weights = np.zeros(len(doc))
    for spans, weight in weighted_spans:
        for start, end in spans:
            char_weights[start:end] += weight
    # TODO - can be much smarter, join spans at least
    # TODO - for longer documents, remove text without active features
    weight_range = max(abs(char_weights.min()), abs(char_weights.max()))
    return ''.join(
        _colorize(c, weight, weight_range)
        for c, weight in zip(doc, char_weights))


def _colorize(token, weight, weight_range):
    alpha = (abs(weight) / weight_range) ** 1.5
    return (
        '<span style="background-color: {color}; opacity: {opacity}" '
        'title="{weight:.3f}"'
        '>{token}</span>'.format(
            color=weight_color(weight, weight_range),
            weight=weight,
            opacity=alpha,
            token=token,
    ))

```

It looks like this:



It'd be nice to have a toggler for these modes.

Also, we should also show bias feature; weight scale should probably take bias in account. I can see two options: add a special element for it (e.g. a large colored rectangle), or use background color for the whole text. There is also a general problem of showing non-text features, if Vectorizer is not the only place features come from; I'm not sure how to show it elegantly.

For colors it makes sense to use HSL/HSLA instead of RGB; css supports it natively.

eli5/templates/weights.html
```diff
@@ -0,0 +1,71 @@
...
+            {% for name, coef in w.pos %}
+                <tr>
+                    <td>{{ name|escape }}</td>
+                    <td>{{ "%+.3f"|format(coef) }}</td>
+                </tr>
+            {% endfor %}
+            {% if w.pos_remaining %}
+                <tr>
+                    <td>&hellip;</td>
+                    <td colspan="2">{{ w.pos_remaining }} more positive</td>
```
jnothman: I tend to prefer seeing the strongest positive and negative coefficients together. In HTML this is particularly feasible as you can use colouring to distinguish them visually. Seeing them listed by polarity emphasised that classification is often just as much about rejecting as accepting a class label; it also allows one to easily see cases where positive and negative coefficients have been learnt that cancel each other out (i.e. similar features with similar coefficients but opposite polarity; model probably under-regularised, etc)

jnothman:
> I've tried the current HTML output; it works nice in standalone files, but in IPython plain-text version looks better to me

I think the main problem above is that the "(7733 more positive features)" is placed in the coefficient column which expands it greatly. You're also not exploiting HTML features such as colouring to indicate relative value.

eli5/templates/weights.html
kmike: @jnothman I don't know what should be a default, but an option and a toggler would be nice to have. Sorting coefficients by absolute value looks useful, and color allows to easily see if a feature is positive or negative, a good point.

We should definitely use more HTML features, not just display a table with coefficients.

lopuhin:
@jnothman @kmike thanks for great suggestions!

I fixed a couple of things:
- colorised weights, using the same scheme as for features highlighting, removed borders and fixed row highlighting
- using HSLA space (actually even HSL, as alpha is always 1.0 at the moment)
- added a small dose of opacity to text features (but a mode with more aggressive opacity application definitely makes sense @kmike )
- added BIAS feature display in highlighted text (for now putting all not found features at the start)




Now next step would be adding more interactivity.

eli5/sklearn/explain_prediction.py
```diff
@@ -105,11 +107,20 @@ def _label(label_id, label):
...
         if proba is not None:
             class_info['proba'] = proba[1]
+        _add_weighted_spans(doc, vec, class_info)
         res['classes'].append(class_info)
 
     return res
 
 
+def _add_weighted_spans(doc, vec, class_info):
+    if isinstance(doc, str) and vec is not None:
```
kmike: six.string_types?

lopuhin:
Hey @kmike this is finally ready for review!

Summary of recent changes:
- added support for highlighting features in text for unhashed features - this required moving their formatting to text and html formatters, and now ``explain_*`` functions can return not only strings as feature names, but also lists of ``{"name": x, "sign": y}``. Only the first (most probable) candidate is displayed in html, rest are hidden in title.
- swapped feature and weight columns: that is better for text formatting and for unhashed features, where it makes sense to have the most probable first and closes to the weights.
- switched most styles to inline styles (all except for ":hover", but I still left that file as we can add scripts that support interactive sorting/toggling and other stuff there), so now github displays notebooks as well as the local jupyter notebook: https://github.com/TeamHG-Memex/eli5/blob/html-formatters/notebooks/explain_text_prediction.ipynb
- added tests

I'll see what extra tests can be added if codecov is still unhappy, but apart from that I don't plan to change anything else here.

eli5/formatters/html.py
```diff
@@ -0,0 +1,158 @@
...
+def _weight_range(weights):
+    """ Max absolute feature for pos and neg weights.
+    """
+    return max([abs(coef) for key in ['pos', 'neg']
+                for _, coef in weights.get(key, [])] or [0])
+
+
+def _format_unhashed_feature(feature):
+    """ Format unhashed feature: show first (most probable) candidate,
+    display other candidates in title attribute.
```
kmike: Hm, what do you think about adding it more visible somehow, e.g. adding some symbol to indicate there are known collisions for the feature?

kmike: ah, there is already ellipses, sorry for noise :)

kmike:
Thanks @lopuhin, great work! 
There is a lot of things which can be added or tweaked, but this is an awesome start.

lopuhin:
Thanks @kmike ! :)

PR #23, opened 2016-10-18
Support feature unhashing when explaining regression predictions

lopuhin:
Adding a small helper that handles unhashing and coef_scale.

Test will need a small tweak to work with #21, I'll do it here or in #21.

### DIFF
--- a/eli5/sklearn/explain_prediction.py
+++ b/eli5/sklearn/explain_prediction.py
@@ -168,7 +168,7 @@ def explain_prediction_linear_regressor(
     }
 
     def _weights(label_id):
-        coef = get_coef(clf, label_id)
+        coef = get_coef(clf, label_id, scale=coef_scale)
         scores = _multiply(x, coef)
         return get_top_features_dict(feature_names, scores, top)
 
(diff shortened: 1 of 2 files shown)


eli5/sklearn/explain_prediction.py
```diff
@@ -143,9 +149,10 @@ def _get_X(doc, vec=None, vectorized=False):
 @explain_prediction.register(SGDRegressor)
 def explain_prediction_linear_regressor(
         clf, doc, vec=None, top=_TOP, target_names=None,
-        feature_names=None, vectorized=False):
+        feature_names=None, vectorized=False, coef_scale=None):
     """ Explain prediction of a linear regressor. """
-    feature_names = get_feature_names(clf, vec, feature_names=feature_names)
+    vec, feature_names, coef_scale = _handle_vec(
```
kmike: `coef_scale` variable is unused (see `_weights` function);

lopuhin: Right, oh my! Thanks for the catch @kmike, fixed now.
But the score looks suspicious for some examples, I want to check if there could be something else here.

kmike:
Thanks @lopuhin!

Issue #14, opened 2016-09-26
defer generating dummy feature names in FeatureUnhasher


kmike:
I think FeatureUnhasher.get_feature_names should have an option to use nan / None as feature names instead of generated `FEATURE[%d]` string names. Creating all these string is the slowest part of this code, and it looks unnecessary because printing/formatting code can easily generate missing feature names itself.

PR #27, opened 2016-10-20
Make feature_names include bias term again

lopuhin:
Sorry for changing my ming about this issue again.

Internally, we need to have the bias name in feature_names, and if we always add it to externally passed feature_names, it can be a major bottleneck. This change allows pre-computing feature_names and passing it as a numpy array.

For example, this is ``explain_prediction`` (with pre-computing ``feature_names`` included) for 650 links before http://vmprof.com/#/5b1da0f7cfa4c94cd035fe436b6024f5?id=0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0&view=flames and after http://vmprof.com/#/a126da42ad0e64372b35083c366c0855?id=0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0&view=flames this change. Adding a bias to a long list of feature names and converting it to numpy array took 2/3 of total runtime.

It's also possible to accept ``feature_names`` both with and without bias feature name, adding it if needed. It makes API easier to use, but might lead to unnoticed off-by-one errors. Not sure which way is better?

### DIFF
--- a/eli5/sklearn/explain_prediction.py
+++ b/eli5/sklearn/explain_prediction.py
@@ -29,6 +29,7 @@ from eli5.sklearn.utils import (
     is_probabilistic_classifier,
     has_intercept,
     rename_label,
+    with_bias_name,
 )
 from eli5.sklearn.text import get_weighted_spans
 from eli5._feature_weights import get_top_features_dict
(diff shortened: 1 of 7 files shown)


kmike:
Hm, can it be solved by not converting feature names to np.array?

lopuhin:
Ah, it might be better, let me check this way

lopuhin:
The number of features here is about 2 million (which is 2x * default number of features in a hashing vectorizer). Avoiding making a numpy array and just working with list can make it 4x faster, which is good but still very noticeable (most time is spent in list creation which we can not avoid since we want to modify it by adding a bias feature name).

Another option would be to make feature names sparce (for example a dictionary) and have entries only for non-generated feature names (so solving #14 too). But maybe even better is to just do not add the bias term here, we can carry it until we really need it's name - I'll try that approach.

Issue #30, opened 2016-10-21
Text highlighting: should we preserve density?


lopuhin:
When highlighting a feature, we can highlight it regardless of length (currently in master), or try to preserve density, so coloring longer feature with a less intense color. I tried that second approach in preserve-density branch, here are some screenshots with master behaviour on top (links to notebooks: https://github.com/TeamHG-Memex/eli5/blob/preserve-density/notebooks/explain_text_prediction.ipynb for words and https://github.com/TeamHG-Memex/eli5/blob/preserve-density/notebooks/explain_text_prediction_char.ipynb for chars).

![2016-10-21 11 28 54](https://cloud.githubusercontent.com/assets/424613/19592008/94205cf6-9782-11e6-8ceb-16f1b746fa76.png)

![2016-10-21 11 31 22](https://cloud.githubusercontent.com/assets/424613/19592013/98244ede-9782-11e6-84b5-ce63bd1dfe70.png)

kmike:
I don't see a big difference in these examples. For tokens it made it worse. For chars it probably made it more correct. I think it is a good option to have, +1 to add it and enable for char ngrams.

kmike:
For char_wb and for token ngrams it is not clear what is better.

lopuhin:
I agree that it's worse for tokens, but even for char ngrams, the main difference seems to be that bias (the largest feature) is less intense compared to shorter char n-grams, so they look brighter (about the same brightness as bias).
I'd rather have it off by default in all cases, but I'm not entirely sure.
One more questions: I just realised that currently weights that are displayed on mouse over in title are different depending on the preserve_density option value, that is clearly wrong, we should have original weights in title, right?

kmike:
Let's take 'software' token in the following text as an example:

> <BIAS> hi there, i am here looking for some help. my friend is a interior decor designer. he is from thailand. he is trying to find some graphics software on pc. any suggestion on which software to buy,where to buy and how much it costs ? he likes the most sophisticated software(the more features it has,the better)

In char-based example it is light-green. The total weight it gives to the result is quite large though: about 0.1 (a bit less for some of them) for each of char 3- and 4-grams (' sof', 'sof', 'soft', 'oft', 'oftw', 'ftw', 'ftwa', 'twa', 'twar', 'war', 'ware', 'are'); it sums up to a value larger than BIAS. But bias is much brighter here:

[...]

PR #32, opened 2016-10-21
Preserve density - see #30

lopuhin:
By default density is preserved (new behaviour) for char and char_wb analyzers, and old behaviour is kept for word analyzer, as discussed in #30.

### DIFF
--- a/eli5/formatters/features.py
+++ b/eli5/formatters/features.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 
-class EscapedFeatureName(object):
+class FormattedFeatureName(object):
     """ Feature name that does not need any additional escaping.
     """
     def __init__(self, value):
@@ -11,7 +11,7 @@ class EscapedFeatureName(object):
         return str(self.value)
 
     def __eq__(self, other):
-        return (isinstance(other, EscapedFeatureName) and
+        return (isinstance(other, FormattedFeatureName) and
                 self.value == other.value)
 
     def __hash__(self):
--- a/eli5/formatters/features.py
+++ b/eli5/formatters/features.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+
+class EscapedFeatureName(object):
+    """ Feature name that does not need any additional escaping.
+    """
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return str(self.value)
+
+    def __eq__(self, other):
+        return str(self) == other
+
+    def __repr__(self):
+        return '<{} {!r}>'.format(self.__class__.__name__, self.value)
--- a/eli5/formatters/features.py
+++ b/eli5/formatters/features.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import six
+
 
 class FormattedFeatureName(object):
     """ Feature name that does not need any additional escaping.
@@ -7,8 +9,8 @@ class FormattedFeatureName(object):
     def __init__(self, value):
         self.value = value
 
-    def __str__(self):
-        return str(self.value)
+    def format(self):
+        return six.text_type(self.value)
 
     def __eq__(self, other):
         return (isinstance(other, FormattedFeatureName) and
--- a/eli5/formatters/features.py
+++ b/eli5/formatters/features.py
@@ -11,7 +11,11 @@ class EscapedFeatureName(object):
         return str(self.value)
 
     def __eq__(self, other):
-        return str(self) == other
+        return (isinstance(other, EscapedFeatureName) and
+                self.value == other.value)
+
+    def __hash__(self):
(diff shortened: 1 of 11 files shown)


lopuhin:
Hm, I re-triggered the travis build by squashing commits, but the codecov is still showing low coverage of the patch.

kmike:
Is it rebased on master?

lopuhin:
It was rebased against an old master, I forgot to pull - rebased against up-to-date master now, thanks for the hint!

lopuhin:
Hm, maybe codecov is correct after all :)

lopuhin:
Yeah, it was really lack of tests on my part, fixed now.

eli5/formatters/html.py
```diff
@@ -49,19 +51,29 @@ def format_html_styles():
...
     for _, spans, weight in weighted_spans:
         for start, end in spans:
+            if preserve_density:
+                weight /= (end - start)
             char_weights[start:end] += weight
     # TODO - can be much smarter, join spans at least
     # TODO - for longer documents, remove text without active features
     not_found_weights = sorted(
-        (feature, weight)
+        (feature, weight / len(feature) if preserve_density else weight)
```
kmike: This part is tricky - non-text feature names can be lighter or darker depending on a length of their name; two features with equal weights can have different color if their names have different length. I don't know how to solve it :) My main motivation was to make highlighting compatible between char and token models, not so much between text and non-text features.

Maybe it deserves its own solution - e.g. instead of showing non-text features inline before the text we may show them separated by new lines, and near each feature name show a rectangle, all rectangles of the same size. This way length of non-text feature name won't impact highlighting. To make it compatible with text rectangle may have a width similar to avg. word width (e.g. 5 chars).

lopuhin:
Another thing that does not play nicely with the idea of preserving density is that right now, we take the feature weight that includes the vectorizer score. As a result, if a feature is repeated several times, then its weight might be higher (if using CountVectorizer for example), but it will be highlighted with this higher weight several times. So on the screenshot "space" is more important feature, which is obvious from the first example, but not from the second, where "flight" is repeated several times:


So I think we should evenly divide the weight of a feature among all it's instances, and do it for all analyzers (for words and chars). Even with CountVectorizer(binary=True) that will give the correct impression I think.

eli5/formatters/html.py
lopuhin: > but unfortunately our text highlighting already works only for linear models
Because of adding weights for overlapping features

eli5/formatters/features.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+
+class EscapedFeatureName(object):
```
kmike: What do you think about renaming it to something like FormattedFeatureName or PreparedFeatureName? If I understood the code correctly, instances of this class are passed as-is through formatting functions, this is not directly related to HTML escaping.

lopuhin: Yes, I like `FormattedFeatureName`, thanks for the suggestion! Changed it.

eli5/formatters/features.py
```diff
@@ -0,0 +1,21 @@
...
+    """ Feature name that does not need any additional escaping.
+    """
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return str(self.value)
```
lopuhin: Right, I didn't think about Python 2 here, thanks for the catch!
I want it to return text type, so it seems easier to use a `.format()` method instead: ee0c4ae.

tests/test_formatters_text.py
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
-from eli5.formatters import EscapedFeatureName
+from eli5.formatters import FormattedFeatureName
 from eli5.formatters.text import _format_feature, _SPACE
 
 
 def test_format_escaped_feature():
```
kmike: it makes sense to change test names as well

lopuhin: Right, I missed that, thanks! Added to 1a98be8

eli5/formatters/features.py
```diff
@@ -1,14 +1,16 @@
...
 class FormattedFeatureName(object):
     """ Feature name that does not need any additional escaping.
     """
     def __init__(self, value):
         self.value = value
 
-    def __str__(self):
-        return str(self.value)
+    def format(self):
+        return six.text_type(self.value)
```
kmike: What's six.text_type call for? When self.value can be bytes? In Python 2.x `unicode(b'string')` works correctly only for ascii-only byte strings because it tries to decode bytes data using sys.getdefaultencoding(). In Python3 str(b'foo') returns `"b'foo'"`, not `"foo"`.

lopuhin: Fixed now. Didn't think about python 2 string for a while, sorry for the mess :)

PR #33, opened 2016-10-21
Make space visible in html and text formatters

### LINKED ISSUE #31: html features: preserve whitespaces
Features with whitespaces in front get these whitespaces removed in HTML.

Compare:

```
+2.837  spa 
+2.805   spa
```

and 

I think whitespaces should be replaced with `&nbsp;` for HTML display. It could also make sense to use another background for text, in order to show whitespaces in the end.

lopuhin:
Fixes #31


eli5/formatters/utils.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+
+def replace_starting_trailing_spaces(s, replacer):
```
kmike: Why not replace all whitespaces?

lopuhin: Yeah, I see, that could help indeed. I think having spaces both inside and outside should be rather rare, but the user can notice how space is marked when working on bigram features, I'll check how it looks like.
And also let's add a title to that span with some message that tells that this is a space here, like "A space symbol"

kmike: Good idea

lopuhin: Spaces within features also look fine, so I added highlighting of all spaces and the title attribute:

kmike: is `replace_starting_trailing_spaces` name still correct?

lopuhin: Right, it handles all spaces now, thanks for the catch! Fixed.

kmike:
Thanks @lopuhin!

PR #34, opened 2016-10-24
allow explanations to be less verbose

kmike:
I've added 'show' argument to format_as_... methods which allow to set which fields to display and their order. What do you think @lopuhin?


lopuhin:
I like that @kmike , I was already stripping unneeded text in a hacky way in one of the notebooks. Thanks!

PR #37, opened 2016-10-31
allow to print DecisionTreeClassifier as text or svg

kmike:
Currently output looks like this:

```
petal width <= 2.450  (33.3%)  ---> [1.000, 0.000, 0.000]
petal width > 2.450  (66.7%)
    petal length <= 1.750  (36.0%)
        petal width <= 4.950  (32.0%)  ---> [0.000, 0.979, 0.021]
        petal width > 4.950  (4.0%)  ---> [0.000, 0.333, 0.667]
    petal length > 1.750  (30.7%)
        petal width <= 4.850  (2.0%)  ---> [0.000, 0.333, 0.667]
        petal width > 4.850  (28.7%)  ---> [0.000, 0.000, 1.000]
```

### DIFF
--- a/eli5/sklearn/explain_weights.py
+++ b/eli5/sklearn/explain_weights.py
@@ -28,11 +28,13 @@ from sklearn.ensemble import (
     RandomForestClassifier,
     ExtraTreesClassifier
 )
-from sklearn.tree import DecisionTreeClassifier
+from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
+from sklearn.tree.export import export_graphviz
 
 from eli5._feature_weights import get_top_features_dict
 from eli5.utils import argsort_k_largest
 from eli5.sklearn.unhashing import handle_hashing_vec, is_invhashing
+from eli5.sklearn.treeinspect import tree2dict
 from eli5.sklearn.utils import (
     get_coef,
     is_multiclass_classifier,
@@ -42,7 +44,6 @@ from eli5.sklearn.utils import (
     rename_label,
 )
 from eli5.explain import explain_weights
-
 
 
 LINEAR_CAVEATS = """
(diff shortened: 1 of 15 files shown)


eli5/sklearn/explain_weights.py
```diff
@@ -28,11 +28,13 @@
     RandomForestClassifier,
     ExtraTreesClassifier
 )
-from sklearn.tree import DecisionTreeClassifier
+from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
+from sklearn.tree.export import export_graphviz
```
lopuhin: This is not used? Didn't know about that function!

eli5/sklearn/treeinspect.py
```diff
@@ -0,0 +1,68 @@
...
+        if not node['is_leaf']:
+            feat_id = node['feature_id']
+            if feature_names is None:
+                node['feature_name'] = "x%s" % feat_id
+            else:
+                node['feature_name'] = feature_names[feat_id]
+
+
+def _node2dict(tree, node_id):
+    is_leaf = tree.children_left[node_id] == _tree.TREE_LEAF
```
lopuhin: can a left child be a leaf while right non-leaf?

lopuhin:
Cool, that's much more informative than just feature weights!

eli5/sklearn/explain_weights.py
kmike: yeah, it is unused; I want to use it for HTML export

eli5/sklearn/treeinspect.py
kmike: the code above is a way to check if a current node is a leaf, not if its left child is a leaf; I copy-pasted it from export_graphviz

lopuhin: Right, now I understand, thanks!

PR #45, opened 2016-11-02
Use attrs-based objects to hold results of explain_weights and explain_prediction

lopuhin:
This is a purely mechanical change apart from one thing: we had distinct "classes" and "targets" keys for classification and regression, and "class"/"target" keys inside them - in this branch I changed to use "targets" in all cases, because it seems applicable to classification too, and simplified code and API (this is similar to what we did with "estimator" key).

Right now I think that changing from dicts to objects is a good thing to do, because of:
- the API will be more future-proof - we can deprecate stuff, add properties, etc.
- potentially, IDE support is better (~~right now PyCharm does not understand attrs classes, but we can fix it by adding ``__init__`` to them~~, and also add type annotations)
- we can also have more benefit from mypy-lang and typing in the future
- we can catch errors where we mistype an attribute name more easily (here we declare up-front what attributes does an object have)

We can still convert objects to dicts using ``attr.asdict`` function (it is recursive by default).

The only downside I see so far is that ``__repr__`` of attrs objects contains the class name and all attribute names and values, but they are not formatted as nicely as IPython can do it (there are no line breaks added). But we can disable this custom ``__repr__`` or add our own.

There are no new features here - the branch touches a lot of code, so to minimize conflicts it will be easier to merge it first and then add new features.

### DIFF
--- a/eli5/base.py
+++ b/eli5/base.py
@@ -14,7 +14,6 @@ class Explanation(object):
     decision_tree = attr.ib(default=None)
 
 
-# TODO - split into two classes?
 @attr.s
 class TargetExplanation(object):
     target = attr.ib()
@@ -28,8 +27,8 @@ class TargetExplanation(object):
 class FeatureWeights(object):
     pos = attr.ib()
     neg = attr.ib()
-    pos_remaining = attr.ib()
-    neg_remaining = attr.ib()
+    pos_remaining = attr.ib(default=0)
+    neg_remaining = attr.ib(default=0)
 
 
 @attr.s
@@ -37,4 +36,4 @@ class WeightedSpans(object):
     analyzer = attr.ib()
     document = attr.ib()
     weighted_spans = attr.ib()
-    other = attr.ib()
+    other = attr.ib(default=None)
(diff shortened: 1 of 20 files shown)


lopuhin:
Forgot to say - I additionally checked html output with master, there were only a couple of differences in the number of negative and positive features in newsgroup tests, probably there is still some non-deterministic behaviour left in scikit.

eli5/base.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+
+import attr
+from typing import Dict, List, Tuple, Union
```
kmike: what do you think about swapping imports? I always think about typing as an stdlib module

eli5/base.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@attr.s
+class Explanation(object):
```
kmike: Could you please add docstrings to all these classes? They don't have to be long; maybe you can move some of comments from explain_.. functions here as well.

lopuhin: Right, docstrings and some comment would be definitely nice here, let me add them tomorrow :)

kmike: Let's do it in another PR; this one looks good as-is!

PR #47, opened 2016-11-02
Format explanation as html in the notebook

lopuhin:
This way result of explain_weights and explain_prediction is shown with html formatter in IPython. Almost all styles are inline anyway, but there is a tiny bit that is included with every output. I think this is not a problem at the moment, but I don't know how to solve it in general without awful hacks.

### DIFF
--- a/eli5/_feature_weights.py
+++ b/eli5/_feature_weights.py
@@ -2,10 +2,11 @@
 from __future__ import absolute_import
 import numpy as np
 
+from eli5.base import FeatureWeights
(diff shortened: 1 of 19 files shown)


kmike:
What do you think about using show=fields.WEIGHTS here?

lopuhin:
Yes, that makes perfect sense to make the explanation more compact and informative, thanks!

lopuhin:
git won this round, continued in #49

PR #55, opened 2016-11-03
TST check types on Travis

kmike:
Currently not a lot of things is checked because `mypy --silent-imports --check-untyped-defs eli5` fails with multiple errors. There is a lot of errors like this when `--check-untyped-defs` option is enabled:

```
eli5/_feature_weights.py: note: In function "get_top_features":
eli5/_feature_weights.py:41: error: Unexpected keyword argument "pos" for "FeatureWeights"
eli5/_feature_weights.py:41: error: Unexpected keyword argument "neg" for "FeatureWeights"
eli5/_feature_weights.py:41: error: Unexpected keyword argument "pos_remaining" for "FeatureWeights"
eli5/_feature_weights.py:41: error: Unexpected keyword argument "neg_remaining" for "FeatureWeights"
```

but at least some checks are good; at least it will keep type declarations in sync (they were broken).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
     - TOXENV=py27
     - TOXENV=py35
     - TOXENV=py35-nodeps
+    - TOXENV=mypy
 
 addons:
     apt:
--- a/eli5/lime/textutils.py
+++ b/eli5/lime/textutils.py
@@ -4,7 +4,7 @@ Utilities for text generation.
 """
 from __future__ import absolute_import
 import re
-from typing import List, Tuple
+from typing import List, Tuple, Any
 
 import numpy as np
 from sklearn.utils import check_random_state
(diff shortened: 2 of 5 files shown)


lopuhin:
That's great! 👍 
And maybe we can fix some type errors with ``--check-untyped-defs`` by writing ``__init__`` methods for attr-based classes.

lopuhin:
Wow, mypy with ``--check-untyped-defs`` caught an actual error!
```
eli5/sklearn/utils.py: note: In function "get_feature_names":
eli5/sklearn/utils.py:99: error: Argument 1 to "len" has incompatible type "FeatureNames"; expected "Sized"
```
where I copied a piece of code and did not change it (the fix is https://github.com/TeamHG-Memex/eli5/commit/6abbb8db0559757fdd127e2b1e09bd24a3cc60a6#diff-47f7891fe007f47de0c660890504d78c)

@kmike ``--check-untyped-defs`` looks useful, what do you think about adding explicit ``__init__`` methods that enumerate all arguments and moving type declarations to them? Maybe it's also possible to check that they are in sync with definitions in the class fields automatically.

I'm +1 to merge this PR as-is, just discussing future steps :)

kmike:
@lopuhin I may have broken your branch because I rebased "mypy" branch, sorry!

lopuhin:
@kmike np, I was going to redo it anyway!

PR #60, opened 2016-11-07
Mypy check untyped

lopuhin:
Enable checking untyped defs with mypy.

In order to do that, I had to add ``__init__`` methods to base classes. To slightly avoid repetition, I removed explicit ``attr.id`` lines and instead add them based on ``__init__`` signature - I don't see obvious downsides here or ways it could be misused (this is a bit magical, but ``attr.ib`` is also magical), but still that might cause problems or confusion - if so, we can roll back the attributes, while having ``__init__`` methods too.

### DIFF
--- a/eli5/_feature_weights.py
+++ b/eli5/_feature_weights.py
@@ -24,7 +24,7 @@ def _get_top_features(feature_names, coef, top):
       no more than ``num_neg`` negative features.
     """
     if isinstance(top, (list, tuple)):
-        num_pos, num_neg = top
+        num_pos, num_neg = list(top)  # "list" is just for mypy
         pos = _get_top_positive_features(feature_names, coef, num_pos)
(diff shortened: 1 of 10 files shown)


kmike:
See also: https://github.com/hynek/attrs/issues/97

lopuhin:
Thanks! Looks like https://github.com/python/mypy/issues/1310 is a blocker for https://github.com/hynek/attrs/issues/97, and the mypy issue looks very far away, sadly.

I though if this trick with generating attributes from ``__init__`` could be useful for attrs, but it seems to be a bit too hacky to expose in such a base library.

kmike:
Yeah, that mypy issue is a blocker. Also, it proposes a plugin architecture, but this means fix would be mypy-specific and other users (eg. PyCharm) won't benefit from that.

Issue #38, opened 2016-11-01
don't highlight whitespace as red for zero weights


kmike:
It looks weird with feature imporances:

![2016-11-01 18 39 35](https://cloud.githubusercontent.com/assets/107893/19891578/f09fd7c2-a06a-11e6-83f6-385d49e245f0.png)

lopuhin:
It also looks weird for non-text feature names, maybe we should highlight spaces only when we are sure that features are from text, or if the user ask for it.

kmike:
@lopuhin makes sense.

kmike:
I think that it is fixed by #61 for practical purposes.

PR #61, opened 2016-11-07
Highlight spaces conditionally

lopuhin:
This should fix a part of #38 - now we'll only highlight spaces if ``highlight_spaces`` is set to True, or if there are any "invisible" spaces in the explanation.
It is also possible to force space highlighting if we are passed a text vectorizer in explain_weights or explain_prediction - do you think it's worth doing @kmike ?

### DIFF
--- a/eli5/base.py
+++ b/eli5/base.py
@@ -23,6 +23,7 @@ class Explanation(object):
                  targets=None,  # type: List[TargetExplanation]
                  feature_importances=None,  # type: FeatureWeights
                  decision_tree=None,  # type: TreeInfo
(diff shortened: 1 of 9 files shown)


kmike:
> It is also possible to force space highlighting if we are passed a text vectorizer in explain_weights or explain_prediction - do you think it's worth doing @kmike ?

I don't have a strong opinion. What's your preference?

lopuhin:
Thanks!

> I don't have a strong opinion. What's your preference?

I'd postpone doing it - it's not obvious that it is needed now.

PR #62, opened 2016-11-08
Multitarget single table

lopuhin:
Here is an example notebook: https://github.com/TeamHG-Memex/eli5/blob/multitarget-single-table/notebooks/explain_text_prediction.ipynb

When ``dense_multitarget=True`` is passed, feature weights for multiple targets are shown side-by-side in separate tables (that are put inside one big table) with global weight range. Wrapping tables (instead of rendering it as one table) makes code easier, and seems to work fine.
As for feature highlighting in text, the best idea I came up with was to show text only for the winning prediction. Maybe you had something else in mind here @kmike ?


kmike:
Horizontal layout looks awesome, much easier to read!

kmike:
For in-text highlighting I was also thinking about highlighting different words in different colors, color=class. I don't know how well would it work. Maybe bars of different colors for each words. Or maybe color of the top class for this word, with intensity proportional to its difference from weights of other classes. Or maybe this is all pointless, and interactive example could be simpler and easier to implement - just allow to click on a class to highlight text according to weights for this class.

eli5/formatters/html.py
```diff
@@ -39,8 +43,10 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
     in the document.
     If ``highlight_spaces`` is None (default), spaces will be highlighted in
     feature names only if there are any spaces at the start or at the end of the
-    feature. Setting it to True forces space highlighting, and setting it to False
-    turns it off.
+    feature. Setting it to True forces space highlighting, and setting it to
+    False turns it off.
+    ``dense_multitarget`` set to True displays a multitarget classifier weights
+    in a single table, and shows highlighted document only for winning target.
```
kmike: How does it work for regression tasks? I'm not sure it makes sense to choose a 'winner' in multitarget regression.

lopuhin: Right, that does not make any sense for regression, I didn't realise that.

kmike: What do you think about having two separate options:

1) vertical vs horizontal layout of weight tables;
2) how to highlight the text and which examples to show?

kmike: Yeah, it doesn't have to be a single PR; horizontal weights layout is already a nice improvement.

lopuhin:
@kmike I left only the horizontal layout option here, renamed it to ``horizontal_layout`` and made the default.

eli5/formatters/html.py
```diff
@@ -40,8 +41,10 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
     in the document.
     If ``highlight_spaces`` is None (default), spaces will be highlighted in
     feature names only if there are any spaces at the start or at the end of the
-    feature. Setting it to True forces space highlighting, and setting it to False
-    turns it off.
+    feature. Setting it to True forces space highlighting, and setting it to
+    False turns it off.
+    If ``horizontal_multitarget`` is True (default), multitarget classifier
```
kmike: the option is named "horizontal_layout". I'd also use "multiclass" term instead of "multitarget" because it is more common for classifiers.

eli5/formatters/html.py
```diff
@@ -56,9 +59,17 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
         td1_styles='padding: 0 1em 0 0.5em; text-align: right; border: none;',
         tdm_styles='padding: 0 0.5em 0 0.5em; text-align: center; border: none;',
         td2_styles='padding: 0 0.5em 0 0.5em; text-align: left; border: none;',
+        dense_multitarget_table_styles=
+        'border-collapse: collapse; border: none;',
+        dense_multitarget_td_styles='padding: 0px; border: 1px solid black;',
```
kmike: what do you think about renaming these variables as well?

lopuhin: Right, it's better to update them, thanks!

eli5/formatters/html.py
lopuhin: Thanks for the catch, thanks!

kmike:
Thanks @lopuhin!

PR #64, opened 2016-11-09
Require tabulate>=0.7.7 due to showindex use

lopuhin:
"showindex" is not supported in 0.7.5, and 0.7.7 was just released (2 days ago). I noticed because I happened to have 0.7.5 already installed, and tests failed with ``TypeError: tabulate() got an unexpected keyword argument 'showindex'``

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
         'six',
         'typing',
         'graphviz',
-        'tabulate',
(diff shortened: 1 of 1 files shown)


kmike:
Ouch, I also had tabulate 0.7.5 installed, but upgraded it just before writing the code :) Thanks!

PR #66, opened 2016-11-10
vertical-align: top for tds wrapping tables

lopuhin:
In the autologin notebook https://github.com/TeamHG-Memex/autopager/blob/master/notebooks/Training.ipynb I noticed a problem with alignment when there is a different number of features for different targets:

This fixes the problem:


kmike:
Thanks!

PR #67, opened 2016-11-10
Add feature filtering in explain_weights

### LINKED ISSUE #65: allow to filter features by their names
Sometimes it is useful to check coefficients only for some of the features. For example, [here](https://github.com/TeamHG-Memex/autopager/blob/master/notebooks/Training.ipynb) (scroll down to "What are important features?") one may want to check how e.g. query:... features affect the result, without looking at all other features. This also can be helpful when adding a new feature.

What about adding 'feature_re' or 'feature_patterns' argument to explain_weights functions?

lopuhin:
Also move FeatureNames class into it's own module. Fixed #65

I implemented filtering by ``feature_re`` for explain_weights, but got some questions along the way. @kmike what are you thoughts on:
* ``feature_re`` is a string (or a regular expression) that is passed into ``re.search`` - should it be ``re.match`` instead? I always found ``re.match`` less intuitive, but maybe it's just me.
* This feature is not implemented for decision tree - while it's maybe possible to do it somehow, but it's probably not easy, so I'd leave it out for now?
* Do you think it makes sense to add an indication about the number of features filtered out, like we show the number of features not shown? Should it be a separate number (and line)?

### DIFF
--- a/eli5/_feature_names.py
+++ b/eli5/_feature_names.py
@@ -63,8 +63,7 @@ class FeatureNames(object):
             indexed_names = six.iteritems(self.feature_names)
         elif self.feature_names is None:
             indexed_names = []
-        else:
-            assert False, 'should be checked in __init__'
+        assert indexed_names is not None
         if isinstance(feature_re, six.string_types):
             filter_fn = lambda x: re.search(feature_re, x, re.U)
         else:
@@ -74,7 +73,6 @@ class FeatureNames(object):
                 indices.append(idx)
                 filtered_feature_names.append(name)
         return (
-
             FeatureNames(
                 filtered_feature_names,
                 bias_name=self.bias_name,
--- a/eli5/_feature_names.py
+++ b/eli5/_feature_names.py
@@ -1,0 +1,74 @@
+import re
+
+import numpy as np
+
+
+class FeatureNames(object):
+    """
+    A list-like object with feature names. It allows
+    feature names for unknown features to be generated using
+    a provided template, and to avoid making copies of large objects
+    in get_feature_names.
+    """
+    def __init__(self, feature_names=None, bias_name=None,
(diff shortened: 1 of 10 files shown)


kmike:
Hey @lopuhin!

> feature_re is a string (or a regular expression) that is passed into re.search - should it be re.match instead? I always found re.match less intuitive, but maybe it's just me

+1 to use re.search. For me even a simple substring lookup covers 95% cases.

> This feature is not implemented for decision tree - while it's maybe possible to do it somehow, but it's probably not easy, so I'd leave it out for now?

Yeah, it is not clear what to do with decision trees. Let's leave them out for now.

> Do you think it makes sense to add an indication about the number of features filtered out, like we show the number of features not shown? Should it be a separate number (and line)?

It won't harm, but I don't think this number is super-useful.

lopuhin:
>> Do you think it makes sense to add an indication about the number of features filtered out, like we show the number of features not shown? Should it be a separate number (and line)?

> It won't harm, but I don't think this number is super-useful.

Ok, I'm also not sure how useful it is, let's leave it out for now.

I also made this filtering not applicable to feature names generated with ``unkn_template``, so unknown unhashed features and just unnamed features will be always filtered out.

Should be ready for review once CI passes.

kmike:
Could you please also fix it for eli5.sklearn_crfsuite? Probably that's fine to only filter state features.

lopuhin:
> Could you please also fix it for eli5.sklearn_crfsuite? Probably that's fine to only filter state features.

Right, I forgot that it was where you had this idea in the first place! I added filtering to state features.

I also moved feature filtering out of target/class loop for multiclass classifiers to avoid repeated filtering.

kmike:
Awesome, thanks!

PR #68, opened 2016-11-10
Fix return values of placeholder functions

lopuhin:
This allows to get nicer type completions in the IDE.

I forgot to do it when replacing dicts with objects.


kmike:
A nice catch.

PR #74, opened 2016-11-14
Fix notebooks after move to attrs

### LINKED ISSUE #73: some of IPython notebooks are broken
These notebooks got broken after we switched to `@attrs` instead of using dicts for explanation results:

* https://github.com/TeamHG-Memex/eli5/blob/master/notebooks/explain_text_prediction_char.ipynb
* https://github.com/TeamHG-Memex/eli5/blob/master/notebooks/explain_text_prediction_unhashing.ipynb

lopuhin:
Fix #73: update hack to show only winning class.

Fixed notebooks:
- https://github.com/TeamHG-Memex/eli5/blob/post-attrs-notebooks/notebooks/explain_text_prediction_char.ipynb
- https://github.com/TeamHG-Memex/eli5/blob/post-attrs-notebooks/notebooks/explain_text_prediction_unhashing.ipynb


kmike:
Thanks, that was quick 👍

PR #90, opened 2016-11-16
Share weights range across all targets

### LINKED ISSUE #88: For multiclass classification weight ranges used for text highlighting are not the same for all classes
On a screenshot the letter 'd' in 'medication' word has -0.084 weight in the top example and -0.079 weight in the bottom example, according to titles displayed on mouse hover. But the bottom 'd' is brighter. Is it because weight range is different? If so, does it make sense to use the same weight range for all classes?

text-highlighting

lopuhin:
Fixes #88
Still there are several separate weight ranges - one for targets, one for text, one for "other" features in text, and one for feature importances.
Weighted spans are passed already rendered now.

@kmike can you please check this on our example in #88 to be extra sure? There are not much tests for this, and they seem hard to write. On newsgroups, new weights range look more intuitive now. I'll update the notebooks one they finish re-evaluating.


kmike:
Thanks @lopuhin! I'll check it after merging :)

lopuhin:
Thanks!

kmike:
It looks much better on newsgroups example, thanks!

PR #91, opened 2016-11-17
Add format_as_dict, remove asdict method

lopuhin:
Apart from consistency with other formatters, another benefit is that it's possible to use format_as_dict on parts of explanation too, which can be useful when sending data to other clients.

### DIFF
--- a/docs/source/autodocs/formatters.rst
+++ b/docs/source/autodocs/formatters.rst
@@ -14,3 +14,9 @@ eli5.formatters.text
 
 .. automodule:: eli5.formatters.text
     :members:
+
+eli5.formatters.as_dict
+-----------------------
+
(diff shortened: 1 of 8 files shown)


kmike:
Could you please add it to generated docs (https://github.com/TeamHG-Memex/eli5/blob/master/docs/source/autodocs/formatters.rst)?

lopuhin:
> Could you please add it to generated docs

Done (messed up formatting in the first commit). I put it last cause it's probably not the first thing you want from formatters, or it's better to keep the lexicographic order?

kmike:
>  I put it last cause it's probably not the first thing you want from formatters, or it's better to keep the lexicographic order?

Makes sense, I think that's perfect.

PR #93, opened 2016-11-18
Apply feature_re to bias too

### LINKED ISSUE #92: Impossible to get both bias and filtered features (using feature_re)
Not sure what is the best solution here - always return the bias even with feature_re, or add an option?

lopuhin:
Fixes #92 - now it's possible to include bias with ``feature_re`` with a regexp like ``(old_regexp|^<BIAS>$)``


kmike:
Thanks!

PR #94, opened 2016-11-18
speedup eli5.explain_.. functions

kmike:
Features with zero coefficients are no longer shown. For one example it reduced explain_prediction time from 20s to 40ms: most coefficients are zero when linear model coefficients are multiplied by instance vector.


lopuhin:
Neat! 👍 
It should mostly speed up explain_prediction, right?

kmike:
Yeah; speedup for explain_weights will only show up if coefficients are sparse.

PR #96, opened 2016-11-21
Highlight features in text for FeatureUnion

lopuhin:
The approach is to highlight features for each of the vectorizers of the FeatureUnion, then render them and merge results.

TODO:

- [x] the bias term is not displayed for some reason
- [x] ~check (and test) support for unhashed features: something might be missing at the moment~ they don't work and I'd like to do it in a separate PR
- [x] tests for FeatureUnion, including cases of mixing text and non-text vectorizers

### DIFF
--- a/eli5/formatters/text_helpers.py
+++ b/eli5/formatters/text_helpers.py
@@ -14,12 +14,11 @@ def get_char_weights(doc_weighted_spans, preserve_density=None):
     If preserve_density is True, then color for longer fragments will be
     less intensive than for shorter fragments, so that "sum" of intensities
     will correspond to feature weight.
-    If preserve_density is None, then it's value is chosen depending on
-    analyzer kind: it is preserved for "char" and "char_wb" analyzers,
-    and not preserved for "word" analyzers.
+    If preserve_density is None, then it's value is taken from
+    the corresponding attribute of doc_weighted_spans.
     """
     if preserve_density is None:
-        preserve_density = doc_weighted_spans.analyzer.startswith('char')
+        preserve_density = doc_weighted_spans.preserve_density
     char_weights = np.zeros(len(doc_weighted_spans.document))
     feature_counts = Counter(f for f, _, _ in doc_weighted_spans.spans)
     for feature, spans, weight in doc_weighted_spans.spans:
--- a/eli5/formatters/text_helpers.py
+++ b/eli5/formatters/text_helpers.py
@@ -34,11 +34,11 @@ def get_char_weights(doc_weighted_spans, preserve_density=None):
 @attrs
 class PreparedWeightedSpans(object):
     def __init__(self,
-                 weighted_spans,  # type: WeightedSpans
+                 doc_weighted_spans,  # type: DocWeightedSpans
                  char_weights,  # type: np.ndarray
                  weight_range,  # type: float
                  ):
-        self.weighted_spans = weighted_spans
+        self.doc_weighted_spans = doc_weighted_spans
         self.char_weights = char_weights
         self.weight_range = weight_range
 
@@ -46,8 +46,8 @@ class PreparedWeightedSpans(object):
         # Working around __eq__ on numpy arrays
         if self.__class__ == other.__class__:
             return (
-                (self.weighted_spans, self.weight_range) ==
-                (other.weighted_spans, other.weight_range) and
+                (self.doc_weighted_spans, self.weight_range) ==
+                (other.doc_weighted_spans, other.weight_range) and
                 self.char_weights.shape == other.char_weights.shape and
                 np.allclose(self.char_weights, other.char_weights))
         return False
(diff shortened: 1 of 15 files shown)


kmike:
Hm, can we move span merging to explanations instead of having it in formatters?

lopuhin:
> Hm, can we move span merging to explanations instead of having it in formatters?

Yes, that's what I originally did, but it had two drawbacks that lead me to move it to formatters:
- the code to merge highlightings is slightly more complex (but it's just 5 lines to re-calculate span starts and ends, so it's not a problem)
- right now, I use some html features for better display of merged highlitings - ``<br/>`` to separate lines, ``<b>`` to make vectorizer name bold



But yes, I see the point that it's better to have less code and logic in the formatters. Maybe it's possible to do most merging outside of them, and just have a list of (vectorizer name, document, spans) that are later passed to the formatters.

kmike:
I think the tests could be the following: 

1) can we make it easier to create a formatter which takes JSON output and renders it on a client?
2) can we make it easier to highlight text if vectorizers are not from scikit-learn?

It looks like by moving code to formatters you helped (2), but not (1), while having this code directly in explain_.. functions will help (1), but not (2).

lopuhin:
Right, I forgot about (2). I think it's not a huge problem, most code should be extractable into helpers.
I'll try to refactor the code to have more logic in the explanation part, and to make it generally less confusing :)

kmike:
+1 to helpers; they can be easier to test as well

lopuhin:
I think I'm finally more or less happy with how it works and how it is structured, still need to update unit tests after last refactoring, update docstrings and clean up stuff.

eli5/formatters/text_helpers.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def get_char_weights(doc_weighted_spans, preserve_density=None):
+    # type: (DocWeightedSpans, bool) -> np.ndarray
+    """ Return character weights for a text document with highlighted features.
+    If preserve_density is True, then color for longer fragments will be
+    less intensive than for shorter fragments, so that "sum" of intensities
+    will correspond to feature weight.
+    If preserve_density is None, then it's value is chosen depending on
+    analyzer kind: it is preserved for "char" and "char_wb" analyzers,
```
kmike: It'd be good for formatters not to be aware of scikit-learn. Maybe it is just a matter of documenting things differently.

lopuhin:
@kmike this is finally ready for review :)

eli5/formatters/utils.py
```diff
@@ -68,3 +68,8 @@ def _has_invisible_spaces(name):
     if isinstance(name, list):
         return any(_has_invisible_spaces(n['name']) for n in name)
     return name.startswith(' ') or name.endswith(' ')
+
+
+def max_or_0(it):
```
kmike: I think eli5.utils is a better place for it; it is a general-purpose function not related to formatters directly. Could you please also add a test for it (a doctest is fine)?

tests/test_sklearn_vectorizers.py
```diff
@@ -159,3 +161,74 @@ def test_explain_regression_hashing_vectorizer(newsgroups_train_binary):
...
+        {'url': 'http://c.com/blog',
+         'text': 'security'},
+        ]
+    ys = [1, 0, 0, 0, 1]
+    vec = FeatureUnion([
+        ('url', vec_cls(preprocessor=lambda x: x['url'],
+                        analyzer='char',
+                        ngram_range=(3, 3))),
+        ('text', vec_cls(preprocessor=lambda x: x['text'])),
+    ])
```
kmike: Is FeatureUnion for the same document supported (e.g. analyzer='char' CountVectorizer and analyzer='word' CountVectorizer at the same time)? This feature could be awesome for LIME.

eli5/formatters/utils.py
lopuhin: Yes, I agree, it's not specific to formatters at all. Moved in 9cba801, thanks!

tests/test_sklearn_vectorizers.py
lopuhin: Yes, but they would be still shown as two separate "documents" (although the same) with features from each of the highlighted using different scales. Is this what you have in mind for LIME, or you would like to have just one "document" with features from both vectorizers applied?

kmike: I think showing it in a same document with the same scale (at least optionally) makes more sense, because it is easier to understand which parts of document are important overall. 

Yeah, this is what I had in mind for LIME - I was thinking about a default classifier to use, and 
a logistic regression trained on both char ngram and word unigram features looked like a baseline which should cover a lot of cases. I don't know how what it takes to implement this though :)

kmike:
Thanks @lopuhin, yet another awesome feature from you!

tests/test_sklearn_vectorizers.py
lopuhin: Yes, I see, thanks for the explanation! think it should be possible to modify an explanation that contains two separate documents into one by concatenating ``span`` fields of ``DocWeightedSpans``.

lopuhin:
Thanks for review @kmike !

PR #100, opened 2016-11-22
slice support for FeatureNames

### LINKED ISSUE #97: FeatureNames doesn't support slicing
`ivec.get_feature_names()[:100]` doesn't work for InvertableHashingVectorizer because FeatureName doesn't handle slice objects.

See also: http://stackoverflow.com/questions/13855288/turn-slice-into-range-in-python; there is also a nice comment

> Does it help that in Python 3 you can slice a range object to get a new range object?

-- given that a feature is not super-important we could decide to support it only in Python 3 :)

lopuhin:
Slicing returns a list or sometimes a slice of the numpy array. Fixes #97


kmike:
Nice, thanks!

PR #101, opened 2016-11-22
better IPython support: eli5.show_weights and eli5.show_prediction

### LINKED ISSUE #69: It is hard to customize formatting in IPython notebook
Currently in order to change formatting options in IPython notebook user has to do something like this:

```
from IPython.display import HTML
expl = explain_weights(clf, vec=fe, top=20)
HTML(format_as_html(expl, highlight_spaces=False, horizontal_layout=False))
```

It'd be nice to reduce it to a one-liner.

kmike:
Fixes #69.

TODO:

* [x] update overview;
* [x] use these functions in IPython notebooks.


kmike:
I haven't switched all notebooks to use eli5.show_.. functions, but that shold be ok for now.

PR #103, opened 2016-11-23
Fix target header in horizontal layout

### LINKED ISSUE #99: Headers for targets are not aligned
A minor issue, but since targets are rendered in separate tables, their headers might get misaligned, for example here: http://eli5.readthedocs.io/en/latest/tutorials/sklearn-text.html

It should be easy to move headers rendering out, to the top-level table.

lopuhin:
Move it to the outer table, so it is aligned across all targets. Fixes #99 - this is how it looks now in case targets don't fit in one line:


kmike:
Thanks!

Issue #39, opened 2016-11-01
LIME: add an example on a real dataset


kmike:

kmike:
Fixed by #110.

PR #110, opened 2016-12-13
improved LIME for text data

### LINKED ISSUE #102: LIME: add support for char-based text classifiers

kmike:
This is a work-in-progress.

TODO:

* [x] tutorial for debugging complex text processing pipelines using LIME;
* [x] cleanup LIME documentation, remove redundancy;
* [x] TextExplainer should handle target_names better when there is no examples of some class in the generated dataset;
* [x] fix test coverage

This PR also fixes #102 and #39.


kmike:
//cc @lopuhin I think this is ready. Do you have any comments?

Tutorial doesn't explain all options - most notably, it doesn't use `position_dependent=True` flag and doesn't explain `rbf_sigma` parameter. I know how they work and what they intend to do, but I don't know when should one use them in practice, so it is not in the tutorial :)

docs/source/_notebooks/text-explainer.rst
```diff
@@ -0,0 +1,2241 @@
+
+TextExplainer: debugging complex text processing pipelines using LIME
+=====================================================================
+
+While eli5 supports many classifiers and preprocessing methods, it can't
+support them all.
+
+If a library is not supported by eli5 directly, or the text processing
+pipeline is too complex for eli5, eli5 can still help - it provides an
+implementation of LIME (Ribeiro et al., 2016) algorithm which allows to
```
lopuhin: I think it would be nice to link to the paper (https://arxiv.org/abs/1602.04938)

lopuhin:
Hey @kmike , I just finished reading the notebook, it looks great - I like that most important stuff is at the start, and you show how it can break.

KL divergence and score of the white-box classifier predictions of the black-box classifier seem to be important to know if explanation should be trusted - does it make sense to include them in explanation output by default?

lopuhin:
> ... does it make sense to include them in explanation output by default?

Hm, but I see it's not really convenient to implement, and it's not obvious that they are needed in the explanation.

The PR looks great! I didn't know about ``# type: (...) ->`` syntax, that is much more convenient.

docs/source/_notebooks/text-explainer.rst
kmike: Yeah, makes sense.

kmike:
I was also thinking about adding scores to the output by default, but it was not straightforward to implement indeed. Adding a custom field just for scores looks like a bit too much; putting it to description is not enough because description is hidden by default.

PR #112, opened 2016-12-14
Show number of remaining feature importances

### LINKED ISSUE #109: Add number of remaining features for feature importances
Just like we have it for feature weights, this could show the number of remaining features with non-zero feature importance - it can still be lower than total number of features, I think.

lopuhin:
Fixes #109 

Example output in html:

and in the terminal:

I also added an extra space (for text formatter) between feature importance and feature name, like we already have for feature weights.

### DIFF
--- a/eli5/_feature_weights.py
+++ b/eli5/_feature_weights.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 import numpy as np
 
 from eli5.base import FeatureWeights, FeatureWeight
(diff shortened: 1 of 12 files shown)


lopuhin:
While trying to satisfy pedantic codecov I discovered a couple of other edge cases, the most major is that now features with zero importance are not shown. But now it's ready :)

kmike:
Looks good, thanks!

Issue #95, opened 2016-11-21
Bad browser performance on large text with highlighted features


lopuhin:
Scrolling in the notebook is noticeably laggy when the text with highlighted features is in sight. Maybe it will help if we merge spans with the same weight (at least in case of word analyzers).

kmike:
Yeah, I also saw performance issues, epecially in IPython notebook. Profile showed that browser was recalculating element heights or something like that. I wonder if that's specific to IPython.

lopuhin:
mostly fixed in #115

PR #115, opened 2016-12-20
Join spans with the same weight to improve browser performance for highlighted text

lopuhin:
It makes browser performance much better in case of word features (see #95). I also updated all notebooks that use text highlighting. Two notebooks to compare browser performance:
- https://github.com/TeamHG-Memex/eli5/blob/master/notebooks/explain_text_prediction.ipynb
- https://github.com/TeamHG-Memex/eli5/blob/7fdd06c9b76f37c6ffbdcf4c14560a18c0591896/notebooks/explain_text_prediction.ipynb

Here the difference is not that big, but it's bigger when text is long and features are more sparse.


eli5/formatters/html.py
```diff
@@ -119,12 +120,14 @@ def render_targets_weighted_spans(targets, preserve_density):
...
+    # TODO - for longer documents, an remove text without active features
     return ''.join(
-        _colorize(token, weight, pws.weight_range)
-        for token, weight in zip(
-            pws.doc_weighted_spans.document, pws.char_weights))
+        _colorize(''.join(t for t, _ in tokens_weights),
+                  weight,
+                  pws.weight_range)
+        for weight, tokens_weights in groupby(
+            zip(pws.doc_weighted_spans.document, pws.char_weights),
```
kmike: We're showing weights with .3f precision; what do you think about grouping spans if they have the same weight after formatting? It would obviously make it coupled with templates, but maybe it could help to reduce span count even more (I'm not sure how much though).

kmike: Without weights in title we can combine spans even more aggresively (color range is even smaller), but weights in title look helpful.

lopuhin: Yeah, that's a good idea! Probably that precision is higher than what is noticeable in colors, I'll try to check that.

kmike: Even if formatted weight precision is higher than noticeable in colors, we still have to show weights in title, so we can't join spans with different formatted weights.

Color range can't be more than 256 (as we're using a single HSL component), and it seems for us it is even less because the range is limited.

lopuhin: Ah, right, I didn't realise we have that 256 limit.

lopuhin: Hm, I think it's a bit trickier than that: while we show weight with just 3 significant figures, it's an "absolute" weight, not relative, and color is chosen relative to weight range. So it can be that both weight and weight range are small. And it's also not quite correct that we always show 3 digits for weights in title.

kmike:
For me the difference in your example notebooks is quite large, thanks for the fix! The file also became ~3x smaller.

eli5/formatters/html.py
kmike: Aha, a good point, so it is more tricky. I've merged the PR because it is already a big improvement, thanks!

Issue #114, opened 2016-12-19
Support explain_prediction for xgboost


lopuhin:
The idea is to do something similar to http://blog.datadive.net/interpreting-random-forests/ - which is used by treeinterpreter (https://github.com/andosa/treeinterpreter/pull/7/).

In order to do that, we need to get the path that was used for prediction. There is a `pred_leaf` argument for ``Booster.predict``: http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.Booster.predict that outputs indices of leafs in all trees. Having this indices, we can recover full paths having the tree structure.
One way to get the tree structure is to use ``get_dump`` method http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.Booster.get_dump which accepts ``dump_format="json"`` (only in xgboost master so far).
I'm not sure if there is an easier way to get it? But it seems these two pieces is all we need for explain_prediction in eli5.

lopuhin:
A drawback of using ``dump_format="json"`` is that it's not released, and installing xgboost from github master is not a one-liner. Probably it's easier to parse text xgboost dump instead?

kmike:
I think it makes sense to do both - use json if it is available and parse text if it is not. XGBoost itself used to parse its text format: https://github.com/dmlc/xgboost/blob/6082184cd1ee831c2ac5e30394cd78b1714cd8c9/python-package/xgboost/core.py#L1088

lopuhin:
Cool, thanks for the pointer! I just implemented a converter from text to json format (https://github.com/TeamHG-Memex/eli5/commit/91c230f8104bbb92b5a918a18c06fa63c1b801cd#diff-ae7c2a29f85d11827eab60b1435ae8f4R55), but their code looks more robust :)

lopuhin:
I think that now explain_prediction is working correctly for xgboost in https://github.com/TeamHG-Memex/eli5/compare/xgboost-prediction branch, and it makes sense, at least for mostly categorical features, like text.
I fear that this approach will not work as well for numerical features, e.g. when it is important that some value lies in an interval. There is also a comment in the original blog post that shows another weakness, that it will fail for a simple XOR tree (weight will be zero for top variable). In practice with xgboost we might have trees with different orderings of variables, but then both variables will have zero weight.
It might be possible to fix that if we track positive and negative contributions of each variable separately: then for XOR each variable will be both a positive and a negative feature, and same for interval example.

[...]

PR #117, opened 2016-12-21
explain_prediction for XGBoost

lopuhin:
This follows an idea from http://blog.datadive.net/interpreting-random-forests/ (see also #114)

I did not check the issues from https://github.com/TeamHG-Memex/eli5/issues/114#issuecomment-268212130 yet, but apart from that it feels ready.

I use some functions from ``eli5.sklearn`` - I think it's fair to use them here because we are supporting explain_prediction for an sklearn wrapper of XGBoost. I just made some of them public. Maybe in the future when we support more different vectorizers, it will make sense to use singledispatch for them too.


lopuhin:
Uh, just discovered an important problem about how missing values are handled.

And also it would be nice to support XGBRegressor too.

lopuhin:
@kmike this is ready for review, sorry for posting it too early - I made everything much more complicated initially.

The sum of all feature weights is equal to the score, as for linear models, and I think produced explanations make sense in all cases (including intervals) except for XOR tree.

For example, in case of intervals (``py.test tests/test_xgboost.py::test_explain_prediction_clf_interval -s``), where it is good to have x0 close to 1, and x1 is irrelevant, the following explanations are produced:

```
[0, 1]
Explained as: decision paths
y=True (probability=0.002, score=-6.481) top features
----------------
  -0.569  <BIAS>
  -0.591  x1    
  -5.321  x0    

[1, 1]
Explained as: decision paths
y=True (probability=0.996, score=5.648) top features
----------------
  +6.567  x0    
  -0.351  x1    
  -0.569  <BIAS>

[2, 1]
Explained as: decision paths
y=True (probability=0.001, score=-6.537) top features
----------------
  -0.490  x1    
  -0.569  <BIAS>
  -5.478  x0    
```

So at x0=0 and x0=2 x0 is a negative feature (the example is classified as 0 because of it), and it is positive at x0=1.

eli5/xgboost.py
```diff
@@ -1,21 +1,41 @@
...
+    # xgboost <= 0.6a2, fixed in https://github.com/dmlc/xgboost/pull/1591
+
+    def xgb_feature_importances(xgb):
+        b = xgb.booster()
+        fs = b.get_fscore()
+        all_features = np.array(
+            [fs.get(f, 0.) for f in b.feature_names], dtype=np.float32)
+        return all_features / all_features.sum()
+
+    XGBRegressor.feature_importances_ = property(xgb_feature_importances)
```
kmike: hm, what do you think about making it an utility function which gets feature importances from XGBRegressor, instead of patching XGBRegressor directly?

lopuhin: Right, it's not great to patch it here. Feature importances are also used to determine the number of features, but it's better to pass them explicitly, I'll fix that, thanks!

eli5/xgboost.py
```diff
@@ -48,3 +68,222 @@ def explain_weights_xgboost(xgb,
...
+    names = xgb.classes_ if isinstance(xgb, XGBClassifier) else ['y']
+    display_names = get_target_display_names(names, target_names, targets)
+
+    scores_weights = prediction_feature_weights(
+        xgb, X, feature_names, missing_idx)
+
+    res = Explanation(
+        estimator=repr(xgb),
+        method='decision paths',
+        targets=[],
```
lopuhin: Yes, thanks for spotting it! We can also mention the problem with XOR tree and that weights sum to score.

eli5/xgboost.py
lopuhin: Fixed in e92334d, now ``xgb_feature_importances`` function is used directly, and ``num_features`` are passed explicitly.

eli5/xgboost.py
lopuhin: @kmike I added description and caveats in a9c692e, does it make sense? Is there anything else we should add?

kmike:
Looks great, thanks @lopuhin!

kmike:
Would you mind updating docs and README as well?

lopuhin:
Thanks! Sure, I'll update them.

kmike:
If you have a bit more time there are two more "pony" requests :) 
First is that it'd be nice to support explain_prediction for sklearn ensebles as well, while you're on it. And second, it'd be great to have a docs chapter which explains the method you're using in more details (similar to how the blog post explains it), with a link to the blog post.

lopuhin:
> Can you get different variables on top by changing the dataset (changing random seed used to generate the dataset)?

Yes, fixing seed gives the same feature weights, but maybe it just means that xgboost is deterministic in this case :)
 
> Maybe we should just say that in xor problems it could work, but it is not reliable if most trees happen to choose the same feature at the top?

Right, I like that! I added this and "single" clarifications in 7fe062e301b2e78451e20b119b713d0b7c7a9aa4, will make another PR soon.

eli5/sklearn/text.py
```diff
@@ -82,8 +96,12 @@ def _get_weighted_spans_from_union(doc, vec_union, feature_weights):
     named_found_features = []
     for vec_name, vec in vec_union.transformer_list:
         vec_prefix = '{}__'.format(vec_name)
-        feature_fn = lambda x: (
-            x[len(vec_prefix):] if x.startswith(vec_prefix) else None)
+
+        def feature_fn(x):
+            if (not isinstance(x, FormattedFeatureName)
```
kmike: Do you recall why is FormattedFeatureName needed? I rewrote the if statement into two statements in https://github.com/TeamHG-Memex/eli5/pull/110/, and coverage shows feature_fn never see FormattedFeatureName in our tests.

lopuhin: Right, there were no test for feature union text highlighting where missing features appeared, I added missing test in https://github.com/TeamHG-Memex/eli5/pull/124

kmike: The additional test showed my refactoring was incorrect, thanks! ;)

PR #122, opened 2016-12-26
Explaining sklearn tree prediction

lopuhin:
Based on the treeinterpreter decision paths idea, this adds support for explaining prediction of ExtraTreesClassifier, ExtraTreesRegressor, GradientBoostingClassifier, GradientBoostingRegressor, RandomForestClassifier, RandomForestRegressor.
AdaBoostClassifier, AdaBoostRegressor are not supported yet: the code runs but produces suspicious results, I would rather check them separately if I won't manage to make them work soon :)


lopuhin:
@kmike I implemented AdaBoostRegressor support but then decided that it's better to leave it out for now, because it selects a tree with median score and uses it for prediction. So we report it's weights and they sum to score, but it's misleading, because only few features will be in this tree, and other trees also influenced the prediction. Another option is to report features from all trees, but it's again misleading. So it requires a ton of special caveats and care.
AdaBoostClassifier has even more complex prediction function that comes in two variants, and though it doesn't have this problem with median estimator, but it will be hard to support it outside of scikit-learn.

So I would rather leave these two estimators out for now. Apart from that, everything is ready.

tests/test_sklearn_explain_prediction.py
```diff
@@ -150,3 +163,114 @@ def test_explain_linear_regression(boston_train, reg):
...
+    docs, y, target_names = newsgroups_train_big
+    vec = CountVectorizer(binary=True, stop_words='english')
+    X = vec.fit_transform(docs)
+    clf.fit(X, y)
+    res = explain_prediction(clf, docs[0], vec=vec, target_names=target_names)
+    check_targets_scores(res)
+    print(format_as_text(res, show=fields.WEIGHTS))
+
+
+def explain_treeinterpreter(clf, xs, feature_names, target_names):
```
kmike: is this function used?

eli5/sklearn/explain_prediction.py
```diff
@@ -193,6 +205,232 @@ def _weights(label_id):
...
+            score=score if score is not None else None,
+            proba=proba[1] if proba is not None else None,
+        )
+        add_weighted_spans(doc, vec, vectorized, target_expl)
+        res.targets.append(target_expl)
+
+    return res
+
+
+@explain_prediction_sklearn.register(AdaBoostRegressor)
```
kmike: AdaBoostRegressor is registered here, but you said it is not supported

lopuhin: Oh, right, I forgot to remove it, thanks for the catch!

tests/test_sklearn_explain_prediction.py
lopuhin: It's not used, I used it during development to compare with treeinterpreter. Probably we should not integrate it into our test suite, and we have checks that sum of all weights is equal to score.

kmike:
Looks great, thanks @lopuhin!

lopuhin:
Thanks @kmike !

PR #123, opened 2016-12-27
[wip] Add an option to expand missing features for explain_prediction

lopuhin:
In some cases it can be interesting to check what features are missing during prediction, because missing feature weight can be higher than the sum of all active features. Here ``expand_missing_features`` option is added.

### DIFF
--- a/eli5/xgboost.py
+++ b/eli5/xgboost.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 import re
 from singledispatch import singledispatch
 import six
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Tuple, Union
 
 import numpy as np
 import scipy.sparse as sp
@@ -138,7 +138,8 @@ def explain_prediction_xgboost(
     names = xgb.classes_ if not is_regression else ['y']
     display_names = get_target_display_names(names, target_names, targets)
 
-    missing_map = {}  # map feature idx to missing feature idx
+    # map feature idx to missing feature idx
+    missing_map = {}  # type: Dict[Union[int, str], int]
     for _, (feature_weights, missing_feature_weights) in scores_weights:
         if expand_missing_features:
             for idx, value in six.iteritems(missing_feature_weights):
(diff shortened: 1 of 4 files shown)


eli5/xgboost.py
```diff
@@ -91,7 +93,9 @@ def explain_prediction_xgboost(
         target_names=None,
         targets=None,
         feature_names=None,
-        vectorized=False):
+        vectorized=False,
+        expand_missing_features=False,
```
kmike: When is it good not to expand these features, i.e. why is it False by default?

kmike: I haven't tried it, so likely you're right :) my point was that if xgboost makes decision based on these features then why shouldn't we show them. Maybe some of them have large weights and showing them is more important than showing features which are not missing. There can be a lot of present features as well; we have "top" and "feature_re" arguments to limit the display.

kmike: hm, we don't have feature_re for explain_prediction methods; maybe we should - it could allow to e.g. show only missing features

lopuhin: Right, that's a valid point! Let me work with it a little more, I also have a fairly limited experience with it :)

lopuhin: Yes, it seems that the only place where it makes sense to collapse hidden features is for text, which is a rare use-case for tree ensembles. So let's rename it to "collapse_missing_features" and disable by default?

kmike: A more general implementation could be the following: 

1. support feature_re for explain_prediction;
2. add an option to show combined weights for filtered out features.

But `collapse_missing_features` also sounds fine to me; it is already implemented, and it is a bit different from feature_re if there are non-missing features with `(missing)` in their names.

lopuhin: We'll also need an ability to provide an inverse of feature_re, or make it accept a function - else we won't be able to filter out missing features, right?
I like an option to show combined weights for filtered out features, it can be useful for all classifiers, not just for xgboost.

kmike: Yeah, a good point. I like 'accept function' solution.

lopuhin:
I think https://github.com/TeamHG-Memex/eli5/pull/123#discussion_r94042895 + #126 is a cleaner approach :)

PR #130, opened 2016-12-30
Show feature value for explain_prediction

### LINKED ISSUE #126: Show feature value for explain_prediction
For numeric features it can be valuable to know the feature value: for example, some feature can be either positive or negative depending on it's value, and it can be made more obvious if we add a third column "Value" to the feature weights table.
On the other hand, it's possible (and rather easy) to implement it outside of eli5, so I'm not sure if it's worth including.

lopuhin:
Fixes #126.
It is shown by default (**edit: it's not shown by default, as before**), and it's possible to pass ``show_feature_values=False`` to ``format_as_text`` and ``format_as_html``. Also the first column of weights table is called ``Contribution`` instead of ``Weight`` for ``explain_prediction``.
I considered using some library for formatting weights table, but could not find any that will allow customizing column alignment (we can have a mix of number and strings in the values column that needs to be right-aligned) and also have nice look, so I added a little function instead.

### DIFF
--- a/eli5/formatters/html.py
+++ b/eli5/formatters/html.py
@@ -11,7 +11,7 @@ from eli5 import _graphviz
 from eli5.base import TargetExplanation
 from eli5.utils import max_or_0
 from .utils import (
-    format_signed, has_any_values_for_weights, replace_spaces,
+    format_signed, format_value, has_any_values_for_weights, replace_spaces,
     should_highlight_spaces)
 from . import fields
 from .features import FormattedFeatureName
@@ -28,7 +28,7 @@ template_env.filters.update(dict(
     remaining_weight_color=lambda ws, w_range, pos_neg:
         format_hsl(remaining_weight_color_hsl(ws, w_range, pos_neg)),
     format_feature=lambda f, w, hl: _format_feature(f, w, hl_spaces=hl),
-    format_value=lambda v: _format_value(v),
+    format_value=format_value,
     format_decision_tree=lambda tree: _format_decision_tree(tree),
 ))
 
@@ -56,7 +56,7 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
     targets = explanation.targets or []
     if len(targets) == 1:
         horizontal_layout = False
-    has_values_for_weights = has_any_values_for_weights(explanation)
+    show_values_for_weights = has_any_values_for_weights(explanation)
 
     rendered_weighted_spans = render_targets_weighted_spans(
         targets, preserve_density)
@@ -93,8 +93,8 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
             for other in weighted_spans_others if other),
         targets_with_weighted_spans=list(
             zip(targets, rendered_weighted_spans, weighted_spans_others)),
-        has_values_for_weights=has_values_for_weights,
-        weights_table_span=3 if has_values_for_weights else 2,
+        show_values_for_weights=show_values_for_weights,
+        weights_table_span=3 if show_values_for_weights else 2,
     )
 
 
@@ -271,15 +271,6 @@ def _format_decision_tree(treedict):
         return _graphviz.dot2svg(treedict.graphviz)
     else:
         return tree2text(treedict)
-
-
-def _format_value(value):
-    if value is None:
-        return ''
-    elif np.isnan(value):
-        return 'Missing'
-    else:
-        return '{:+.3f}'.format(value)
 
 
 def html_escape(text):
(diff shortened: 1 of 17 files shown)


eli5/formatters/text.py
```diff
@@ -15,12 +17,15 @@
...
-def format_as_text(expl, show=fields.ALL, highlight_spaces=None):
+def format_as_text(expl, show=fields.ALL, highlight_spaces=None,
+                   show_feature_values=True):
     """ Format explanation as text.
     If ``highlight_spaces`` is None (default), spaces will be highlighted in
     feature names only if there are any spaces at the start or at the end of the
     feature. Setting it to True forces space highlighting, and setting it to False
     turns it off.
+    If ``show_feature_values`` is True (default), feature values are shown
+    if present.
```
kmike: Could you please add this option to `show_prediction` docs?

kmike:
Hm, do you know why are tests failing?

eli5/formatters/text.py
lopuhin: Right, I forgot about it, added in 00484bb - thanks for the catch!

kmike: Ah, yet another gotcha - it should also be in [FORMAT_KWARGS](https://github.com/TeamHG-Memex/eli5/blob/master/eli5/ipython.py#L10).

lopuhin:
> Hm, do you know why are tests failing?

It's strange they didn't fail for me locally, probably I didn't have lightning installed, and maybe didn't run some of them after the recent changes - should be fixed now :)

eli5/formatters/text.py
lopuhin: Aha, right, thanks! Fixed, now show_* works with show_feature_values

eli5/formatters/html.py
```diff
@@ -27,13 +28,15 @@
...
+    format_value=format_value,
     format_decision_tree=lambda tree: _format_decision_tree(tree),
 ))
 
 
 def format_as_html(explanation, include_styles=True, force_weights=True,
                    show=fields.ALL, preserve_density=None,
-                   highlight_spaces=None, horizontal_layout=True):
+                   highlight_spaces=None, horizontal_layout=True,
+                   show_feature_values=True):
```
kmike: If this is True by default we should also update all IPython notebooks.

By the way, what do you think about keeping it False? Values can be good to see in many cases, but they are also confusing and require more explanation and more classifier knowledge (what's the difference between Value and Contribution?). For text explanations what users get with this option set to true is a table like this above the highlighted text:

table

It could be also good to add title attribute to table header, to show some help on what these columns mean. For example, it is not clear if user should multiply Contribution by Value to get the importance or not.

eli5/formatters/text.py
```diff
@@ -119,17 +135,45 @@ def _targets_lines(explanation, hl_spaces):
             target.target,
             scores)
         lines.append(header)
-        lines.append("-" * (sz + 10))
+
+        if explaining_prediction:
+            table_header = ['Contribution', 'Feature']
```
kmike: 👍 I like "Contribution" name, it is more clear than "Weight"

eli5/formatters/html.py
lopuhin: Yeah I like the idea of having explanations in the header, and I'd also add a little "?" to indicate that there is some help there.

I agree about the default, let's keep it False as it was before, and we can mention that option in "Contribution" help.

lopuhin: I made it False by default, as before, and added explanations for weight and contribution columns here: https://github.com/TeamHG-Memex/eli5/pull/130/commits/f938d762c9ec3b45f3e151e57679b152f6cbe635#diff-1f22eee3046204ba9c246464571ff48bR108, and this is how it looks like:

@kmike do you have suggestions about the text in tooltips?

kmike: @lopuhin nice!

As for Contribution docs, it could be easier to understand if we say explicitly that `contribution = weight * feature value` (is it always the case?)

lopuhin: That's true for linear models, but for tree ensembles it's not: they don't have weights, only feature importances. And even if we somehow could calculate the weights, it won't be true I think.
But we can add "(for linear models, contribution = weight * feature value)", I think it will clarify what is contribution even for non-linear ones.

kmike: Sounds good 👍

lopuhin: Thanks for the idea @kmike! I added clarification in 32519e4

kmike:
Thanks @lopuhin, looks good!

PR #132, opened 2017-01-09
feature_filter in addition to feature_re, support for explain_prediction

lopuhin:
~~This is very much WIP, also missing sklearn tree ensemble explain_prediction support.~~

This branch brings two features:
- support ``feature_filter`` (a function) along with ``feature_re``, so now it's possible to exclude certain features
- support feature filtering for all explain_prediction functions

Also ``eli5.utils.mask`` now handles sparse "vectors".

PR #136, opened 2017-01-10
XGBoost Titanic tutorial

lopuhin:
This branch also contains (most) changes from #132, so the diff is larger than it's really is.
The notebook with the tutorial itself is here: https://github.com/TeamHG-Memex/eli5/blob/xgboost-tutorial-2/notebooks/xgboost-titanic.ipynb.

I also added the csv that I downloaded from kaggle (train set) - there is no clear indication whether it's ok to re-distribute it (although there are plenty of copies of it on github), but it's also available here https://vincentarelbundock.github.io/Rdatasets/datasets.html, and the data itself comes from public sources https://vincentarelbundock.github.io/Rdatasets/doc/datasets/Titanic.html so it seems ok to include it for convenience.

The tutorial is rather light, the goal was just to show how to use xgboost, sklearn and eli5 together. @kmike do you have ideas how it could be extended or changed?


docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+predict: survival.
+
+.. code:: python
+
+    from pprint import pprint
+    from sklearn.utils import shuffle
+    from sklearn.model_selection import train_test_split
+    
+    feature_names = data[0][2:]
+    _all_xs = [dict(zip(feature_names, row[2:])) for row in data[1:]]
```
kmike: Can this code be simplified using csv.DictReader?

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+----------------------------
+
+Let's first build a very simple classifier with ``XGBClassifier`` and
+``sklearn.feature_extraction.DictVectorizer``, and check it's accuracy
+with cross-validation:
+
+.. code:: python
+
+    import warnings
+    warnings.filterwarnings('ignore', category=DeprecationWarning)
```
kmike: I think a comment about the purpose of this .filterwarnings could be nice here ("xgboost <= 0.6a2 shows a warning when used with scikit-learn 0.18+" or something like that)

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+
+.. parsed-literal::
+
+    891 items total, 38.4% true
+    [{'Age': '19',
+      'Cabin': '',
+      'Embarked': 'S',
+      'Fare': '10.1708',
+      'Name': 'Dakic, Mr. Branko',
+      'Parch': '0',
```
kmike: this 0 will be treated as an empty value, is it a problem?

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+
+    Accuracy: 0.823 ± 0.008
+
+
+There are two tricky parts in above code. First is that we pass
+``missing=0`` to ``XGBClassifier``. This tells XGBoost to treat zeros as
+missing values, which is how most scikit-learn vectorizers work. It is
+important both for training and for correct feature visualization.
+
+Second tricky bit is that XGBClassifier has some
```
kmike: I think we should document xgboost version used for the tutorial - comments about xgboost issues could become outdated

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+        
+    
+    
+    
+
+
+
+
+In order to calculate a prediction, XGBoost sums predictions of all it's
+trees (the number of trees is controlled by ``n_estimators`` argumnet
```
kmike: argument

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+independent implementation of this algorithm for XGBoost and most
+scikit-learn tree ensembles.
+
+Here we see that classifier thinks it's good to be a female, but bad to
+travel third class. Some features have "Missing" as value (we are
+passing ``show_feature_values=True`` to view the values): that means
+that the feature was missing, so in this case it's good to not have
+embarked in Southampton.
+
+It's possible to show contributions of only features that are present
```
kmike: "of only features" sounds ungrammatical, but I'm not sure :)

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2468 @@
...
+using ``feature_filter`` argument:
+
+.. code:: python
+
+    no_missing = lambda _, v: v != 0
```
kmike: I feel it is better to use long parameter names here, because users don't know yet what functions feature_filter accepts.

kmike:
The tutorial looks great!

It seems there are still unrelated commits and changes from #132 (it is merged now).

kmike:
I think that's fine to bundle csv file. It won't be obvious that the file is bundled when reading the tutorial from the docs though - we don't say that users should clone the repo and run the notebook from the checkout. Maybe it worths adding a note ("we've downloaded the data file from ... ").

kmike:
Could you please also add a link to this tutorial to http://eli5.readthedocs.io/en/latest/libraries/xgboost.html?

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: I think it's not a problem for classification here: 0 will be treated as "special" value by the decision tree (but this seems ok), and there are no really missing values in "Parch", and zero here means that there are no parents/children abroad.
For visualising predictions this is more problematic, because we filter out missing values.
So in general this is bad, let me think how it can be solved...

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yeah, 100% agree, thanks for spotting that! Fixed in d044c15

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yeah, it sounds strange, I agree! I changed it to "It's possible to show only features that are present using feature_filter argument:" in d044c15

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: right, thanks) Fixed in d044c15

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yes, hopefully :) I added xgboost version at the start of the tutorial with a link to their home page in d044c15

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yep, thanks for the suggestion! Added in d044c15

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Oh, that's really better - I think I always missed it :) Using it in d044c15

lopuhin:
I added the link to tutorial from the xgboost docs and removed extra diff. Apart from tutorial itself, it contains the change to how values are formatted: they don't have a mandatory "+", but weights still do.

Now the only remaining issue is treating 0 as missing values - I'll think how to best solve it here.

lopuhin:
Found that I misunderstood what ``missing`` attribute of XBGClassifier means: it turns out it's only used when converting dense input into internal XGBoost sparse matrix, but not if input is already sparse: https://github.com/dmlc/xgboost/issues/341#issuecomment-141755530

And also it turns out I don't understand yet how to combine sparse features with dense features that can have zeros using scipy sparse matrixes and sklearn.

lopuhin:
I wrote a separate issue for missing values in xgboost: #137 - I think it's a blocker for the tutorial in it's current form, but not a 0.3 blocker I think.

lopuhin:
@kmike I merged #142 into this branch and updated tutorial with correct missing handling: https://github.com/TeamHG-Memex/eli5/blob/xgboost-tutorial-2/notebooks/xgboost-titanic.ipynb

lopuhin:
@kmike so hopefully now it's complete and correct, can you please have another look at it? :)

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2608 @@
...
+        x['Parch'] = int(x['Parch'])
+
+2. Simple XGBoost classifier
+----------------------------
+
+Let's first build a very simple classifier with
+`xbgoost.XGBClassifier <http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier>`__
+and
+`sklearn.feature\_extraction.DictVectorizer <http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.DictVectorizer.html>`__,
+and check it's accuracy with cross-validation:
```
kmike: its

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2608 @@
...
+CSC format, so we add a ``CSCTransformer`` to the pipelne. One may be
+templed to just pass ``dense=True`` to ``DictVectorizer``: after all, in
+this case the matrixes are small. But this is not a great solution,
+because we will loose the ability to distinguish features that are
+missing and features that have zero value.
+
+3. Explaining weights
+---------------------
+
+In order to calculate a prediction, XGBoost sums predictions of all it's
```
kmike: its

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2608 @@
...
+    
+
+
+
+
+We see that now there are a lot of features that come from the *Name*
+field (in fact, a classifier based on *Name* alone gives about 0.79
+accuracy). Name features listed in this way are not very informative,
+they make more sense when we check out predictions. We hide missing
+features here because there are a lot of missing features in text, but
```
kmike: there is a lot

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2608 @@
...
+
+.. code:: python
+
+    from sklearn.pipeline import FeatureUnion
+    from sklearn.feature_extraction.text import CountVectorizer
+    
+    vec2 = FeatureUnion([
+        ('Name', CountVectorizer(
+            analyzer='char_wb',
+            ngram_range=(3, 4),
```
kmike: Could you try it with `(2, 4)` - "Mr" is just two chars, so maybe it could help?

kmike:
This looks great @lopuhin!

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yes, that's reasonable. I just tried (2, 4): the quality is slightly less (although it means nothing: 0.822 ± 0.033 for (2, 4) vs 0.831 ± 0.006 for (3, 4)), it picks more features inside names, but "Mr" is handled by 3 grams too, because it's written as 3 letters: "Mr.". I also tried training a model on names alonee, it's again slightly better with (3, 4) than with (2, 4): 0.776 ± 0.039 for (2, 4) vs 0.788 ± 0.029 for (3, 4).
I think it's too small dataset to reliably check which is better, but predictions from (3, 4) model look a bit better in highlighted text, so I'd rather go with it.

kmike: Ok, thanks for checking it!

kmike: I asked because "Mr' was not highlighted in one of the prediction examples, but likely that's not a problem

docs/source/_notebooks/xgboost-titanic.rst
```diff
@@ -0,0 +1,2608 @@
...
+            return xs.tocsc()
+        def fit(self, *args):
+            return self
+        
+    clf = XGBClassifier()
+    vec = DictVectorizer()
+    pipeline = make_pipeline(vec, CSCTransformer(), clf)
+    
+    def evaluate(_clf):
+        scores = cross_val_score(_clf, all_xs, all_ys, scoring='accuracy')
```
kmike: What do you think about using cv=10? by default it is cv=3

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: wow, I was pretty sure that is wrong, but it's actually correct, thanks for the catch!

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Ah, I didn't notice it, right! This is how it looks with (2, 4):

kmike: How does it look without max_features?

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Yep, that gives more reliable results, thanks for the tip!

docs/source/_notebooks/xgboost-titanic.rst
lopuhin: Hm, without max_features it's almost the same, I expected more changes:

And the gap between (3, 4) and (2, 4) got bigger with cv=10

PR #144, opened 2017-01-12
Fix char_wb spans at the start of the document

### LINKED ISSUE #143: Text features might fail to be highlighted sometimes (?)
For example in the xgboost tutorial if we do not set max_features, this happens:

So nothing is highlighted, but the table shows that there are positive features.

lopuhin:
Fixes #143 

The span start can be -1, because for document "ab"the first 3-gram will be " ab" that starts before the start of the document. Another option would be to fix it in _span_analyzers._char_wb_ngrams. But the ngram is really " ab" and it really starts before the start of the document, so I decided to fix it here, although I'm not 100% sure.


kmike:
A good catch!

PR #151, opened 2017-01-19
Use gain for xgboost feature importance

### LINKED ISSUE #150: xgboost: use gain by default in explain_weights
It seems instead of `b.get_fscore()` we should use `get_score(importance_type='gain')` by default: it is more compatible with scikit-learn, and it should give better results (thanks @rushter and @alexeygrigorev for the suggestion). If we make this change Titanic tutorial should be fixed as well.

It'd be nice to also have an option to use 'weight' and 'cover'. I think it could even make sense to allow showing all of these feature importances at the same time, but it could be another issue.

lopuhin:
Fixes #150


kmike:
Do you know why the second show_weights result is not changed in the Titanic tutorial?

eli5/xgboost.py
```diff
@@ -70,8 +72,19 @@ def explain_weights_xgboost(xgb,
...
+    Parameters
+    ----------
+    importance_type : str, optional
+        A way to get feature importance. Possible values are:
+
+        - 'gain' - the average gain of the feature when it is used in trees
+          (default)
+        - 'weight' - the number of times a feature is used to split the data
+          across all trees
+        - 'cover' - the average coverage of the feature when it is used in trees
```
kmike: there is also docs/libraries/xgboost.rst to updtae

lopuhin:
> Do you know why the second show_weights result is not changed in the Titanic tutorial?

oh, right, I forgot to update the whole notebook, thanks for the catch!

eli5/xgboost.py
lopuhin: oops, thanks! fixed in 86cbfb2

lopuhin:
@kmike now I updated the notebook from scratch (turned out I also didn't save one text edit), fixed in 03f066e

docs/source/libraries/xgboost.rst
```diff
@@ -12,9 +12,18 @@ xgboost >= 0.6a2.
...
 
-:func:`eli5.explain_weights` uses feature importances. For
-XGBClassifer_ and XGBRegressor_ there are no additional supported
-keyword arguments; ``target_names`` and ``target`` arguments are ignored.
+:func:`eli5.explain_weights` uses feature importances. Additional
+arguments for XGBClassifer_ and XGBRegressor_:
+
+* ``importance_type`` is a way to get feature importance. Possible values are:
+
+    - 'gain' - the average gain of the feature when it is used in trees
```
kmike: I'm not 100% sure this is a problem, but github shows it as a quote; this line is indented more than necessary (- could be under the first `)

lopuhin: Aha, thanks for the catch! Not the list looks better in built docs too, without an extra blank line. Finally fixed in https://github.com/TeamHG-Memex/eli5/blob/6409efa/docs/source/libraries/xgboost.rst

kmike:
Thanks!

PR #153, opened 2017-01-20
Add top_targets to explain_prediction

### LINKED ISSUE #149: Limit show_prediction only to few classes
I have 300 target classes and showing explanation of all of them make my browser very slow.
it could be very good if show_prediction could show the explanation of the 3 best predictions and not all the classes (y).
How can I do that?

lopuhin:
Fixes #149.
I think I added it to all explain_prediction functions, and also updated the notebooks that had a hack to show the top target.


eli5/explain.py
```diff
@@ -106,6 +106,14 @@ def explain_prediction(estimator, doc, **kwargs):
 
         This argument may be supported or not, depending on estimator type.
 
+    top_targets : int, optional
+        Number of targets to show. When ``top_targets`` is provided,
+        only specified number of targets with highest scores are shown.
```
kmike: What do you think about supporting negative top_targets? E.g. -1 would mean 1 target with lowest score. I'm not sure it is important for classification, but it could be useful for regression.

lopuhin: 👍  I think it's a great idea, and negative values seem had to confuse with anything else.

eli5/explain.py
```diff
@@ -106,6 +106,14 @@ def explain_prediction(estimator, doc, **kwargs):
 
         This argument may be supported or not, depending on estimator type.
 
+    top_targets : int, optional
```
kmike: It should also be documented in show_prediction

eli5/explain.py
lopuhin: Done in 8b8010f

eli5/explain.py
lopuhin: Eh, I always forget about it! Fixed in 8b8010f

kmike: It is not your fault, there is a lot of duplication in eli5 docs

kmike:
Looks great, thanks!

PR #156, opened 2017-01-24
Fix weighted sum calculation in xgboost._parent_value

### LINKED ISSUE #155: issue about _parent_value for xgboost trees
I have checked through the codes in xgboost.py and became quite confused about this part of codes:
```python
def _parent_value(children):
    """ Value of the parent node: a weighted sum of child values.
    """
    covers = np.array([child['cover'] for child in children])
    covers /= np.sum(covers)
    leafs = np.array([child['leaf'] for child in children])
    return np.mean(leafs * covers)
```
I think `leafs * covers` already represents the weighted sum of child values, because `covers` have already been normalized by `covers /= np.sum(covers)`. Thus, I wonder why should this function return `np.mean(leafs * covers)` instead of `leafs * covers` ?

lopuhin:
As pointed out by @SauceCat, np.sum is needed here, not np.mean.
Also add tests that check that score and proba match.

Fixes #155

TODO:

- [x] think of a way to test it?
- [x] update xgboost tutorial


lopuhin:
@kmike I added a direct test, updated the tutorial and added release notes in 04a05d1 - this is ready for review now :)

kmike:
Thanks @lopuhin for the fix and @SauceCat for the catch!

PR #163, opened 2017-03-03
Handle improperly installed xgboost.

### LINKED ISSUE #162: after install, cannot load eli5 due to xgboost required gcc dylib
After installation, I imported and got this issue:

Library not loaded: /usr/local/opt/gcc/lib/gcc/5/libgomp.1.dylib
  Referenced from: /Users/laam/sentifi/lib/python2.7/site-packages/xgboost/./wrapper/libxgboostwrapper.so
  Reason: image not found

How to fix that? I have gcc installed already in /usr/local/opt/gcc/lib/gcc/6/libgomp.1.dylib

kmike:
Fixes #162. I've checked it the same way as @lopuhin suggested - renamed libxgboost.so file.

### DIFF
--- a/eli5/__init__.py
+++ b/eli5/__init__.py
@@ -38,3 +38,9 @@ try:
 except ImportError:
     # xgboost is not available
     pass
+except Exception as e:
+    if e.__class__.__name__ == 'XGBoostLibraryNotFound':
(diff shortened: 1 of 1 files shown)


lopuhin:
Looks good! 👍

PR #169, opened 2017-04-21
LightGBM support

kmike:
TODO:

* [x] tests for `explain_prediction_lightgbm` on regression task;
* [x] tests for `explain_prediction_lightgbm` on classification task;
* [x] implement `explain_prediction_lightgbm` for multiclass classification;
* [x] tests for explain_prediction_lightgbm on multiclass classification task;
* [x] check missing values handling - LightGBM treats missing values as zeros
* [x] add a proper description to Explanation;
* [x] document LightGBM support;
* [x] ~~check `explain_prediction_lightgbm` with pandas.DataFrame input;~~ postponed.
* [x] update README.
* [x] reduce code duplication

### DIFF
--- a/eli5/__init__.py
+++ b/eli5/__init__.py
@@ -44,3 +44,10 @@ except Exception as e:
         pass
     else:
         raise
+
+
+try:
+    from .lightgbm import explain_weights_lightgbm
+except ImportError:
+    # lightgbm is not available
+    pass
--- a/eli5/__init__.py
+++ b/eli5/__init__.py
@@ -34,7 +34,10 @@ except ImportError as e:
 
 
 try:
-    from .xgboost import explain_weights_xgboost
+    from .xgboost import (
+        explain_weights_xgboost,
+        explain_prediction_xgboost
+    )
 except ImportError:
     # xgboost is not available
     pass
@@ -47,7 +50,10 @@ except Exception as e:
 
 
 try:
-    from .lightgbm import explain_weights_lightgbm
+    from .lightgbm import (
+        explain_weights_lightgbm,
+        explain_prediction_lightgbm
+    )
 except ImportError:
     # lightgbm is not available
     pass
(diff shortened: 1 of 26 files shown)


kmike:
@lopuhin I think it is ready for review, could you please check it?

eli5/lightgbm.py
```diff
@@ -0,0 +1,272 @@
...
+    aliases = {'weight': 'split'}
+    coef = lgb.booster_.feature_importance(
+        importance_type=aliases.get(importance_type, importance_type)
+    )
+    norm = coef.sum()
+    return coef / norm if norm else coef
+
+
+def _compute_node_values(tree_info):
+    """ Add node_value key with an average value for non-leaf nodes """
```
lopuhin: Maybe "expected" would be more precise than "average"?

eli5/sklearn/explain_weights.py
```diff
@@ -249,28 +256,17 @@ def explain_rf_feature_importance(estimator,
...
+    return get_feature_importance_explanation(estimator, vec, coef,
+        coef_std=coef_std,
+        feature_names=feature_names,
+        feature_filter=feature_filter,
+        feature_re=feature_re,
+        top=top,
         description=DESCRIPTION_RANDOM_FOREST,
-        estimator=repr(estimator),
-        method='feature importances',
+        is_regression=isinstance(estimator, RegressorMixin),
```
lopuhin: thanks for spotting that :+1:

eli5/lightgbm.py
kmike: A good catch!

eli5/__init__.py
```diff
@@ -44,3 +44,10 @@
         pass
     else:
         raise
+
+
+try:
+    from .lightgbm import explain_weights_lightgbm
```
kmike: These imports were mostly for singledispatch discovery to work - explain_prediction_xgboost is not there as well. But it indeed makes sense to re-export more of these functions; I'll fix that, thanks for the catch!

PR #170, opened 2017-04-21
mypy update

### LINKED ISSUE #168: Update mypy
I tried updating mypy (and using mypy package instead of mypy-lang), and it's exploding on singledispatch decorators - probably it did not analyze them before.

lopuhin:
Using `mypy` instead of `mypy-lang`. Had to work about some existing mypy issues and filed one new. At least they are all known :)
A couple of fixes were actual bugs in types: missing variants, extra commas, misplaced type annotations. Also mypy started to check None return type more strictly (it's required in type signature and explicit return None is required for Option return type).
Fixes #168

### DIFF
--- a/eli5/lightning.py
+++ b/eli5/lightning.py
@@ -16,7 +16,6 @@ from eli5.sklearn import (
 from eli5.explain import explain_prediction, explain_weights
 
 
-@explain_weights.register(BaseEstimator)
 @singledispatch
 def explain_weights_lightning(estimator, vec=None, top=20, target_names=None,
                               targets=None, feature_names=None,
@@ -28,7 +27,6 @@ def explain_weights_lightning(estimator, vec=None, top=20, target_names=None,
     )
 
 
-@explain_prediction.register(BaseEstimator)
 @singledispatch
 def explain_prediction_lightning(estimator, doc, vec=None, top=None,
                                  target_names=None, targets=None,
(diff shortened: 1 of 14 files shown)


lopuhin:
@kmike this is ready to get merged I think, but perhaps it's better to merge it after #169 , so that you don't have to delay the merge to fix more type errors if they arise.

eli5/lightning.py
```diff
@@ -16,7 +16,6 @@
 from eli5.explain import explain_prediction, explain_weights
 
 
-@explain_weights.register(BaseEstimator)
```
kmike: Does it mean that e.g. for SVRGClassifier `explain_weight_lighting` won't be called, and eli5.explain_weights will be called instead?

lopuhin:
yes - do you think it's ok? I think it's easy to fix if this is undesireable

kmike:
I'm not sure; it seem the only practical difference is that after the change TypeError won't be raised if someone passes an argument unsupported by a library for an estimator which is not implemented for a library.

lopuhin:
Ah, I didn't realise that - let me check if fixing it is not too
cumbersome.
Ср, 26 апр. 2017 г. в 17:55, Mikhail Korobov <notifications@github.com>:

eli5/lightning.py
```diff
@@ -28,7 +27,6 @@ def explain_weights_lightning(estimator, vec=None, top=20, target_names=None,
     )
 
 
-@explain_prediction.register(BaseEstimator)
```
kmike: I think it needs a fix similar to sklearn's, becaue BaseEstimator here is not sklearn's BaseEstimator

lopuhin: Right, thanks for the catch! Should be fixed by 49fb318

kmike:
Hm, it seems coverage got down because `explain_prediction_ovr_sklearn` and `explain_weights_ovr_sklearn` are no longer called.

lopuhin:
Hmm, actually I expected them to be called, maybe I missed something.
Yeah, new mypy+our singledispatch use are not a great fit at the moment, but hopefully it's possible to solve it once and for all.

lopuhin:
The reason why coverage changed is that dispatch was happening not the way I expected:
In master, if we call ``explain_weights(ovr)``, it first gets dispatched to ``explain_weights_sklearn(ovr)``, and only then to ``explain_weights_ovr_sklearn``. But in this branch there is no double dispatch, and ``explain_weights_sklearn`` is never called for ``ovr`` classifier in tests.
Now I'll check how to better fix it.

lopuhin:
@kmike I decided to add an explicit call to explain_*_sklearn for OvR to tests instead of changing how OvR is dispatched - it looks simpler to understand. Hopefully now the coverage will improve.

kmike:
Ok, thanks!

PR #171, opened 2017-04-24
Render spans with zero weights: test and fix

lopuhin:
If all weight are zero, previously a runtime warning was raised, and opacity was "nan" in html:
```
eli5/formatters/html.py:195: RuntimeWarning: invalid value encountered in double_scalars
  rel_weight = abs(weight) / weight_range
<span style="opacity: nan">ab</span>
```
Now if weights and weight range are zero, weights stay at zero.


kmike:
👍 Thanks!

PR #173, opened 2017-04-27
Improve lightning coverage for OvR classifiers

lopuhin:
Same as c6c9f94217e1abaa8a8645858f37a386d4794250

Testing just for one classifier, as they all go via the same function.


kmike:
👍

PR #176, opened 2017-04-28
Support hashing vectorizer inside a union

### LINKED ISSUE #16: Support FeatureUnion for InvertableHashingVectorizer
Just adding features from `.transformer_list`, possibly with prefixes, should be enough

lopuhin:
Also add an ``eli5.sklearn.invert_and_fit`` helper that inverts and fits a vectorizer (even if it's hiding in a union). The fix in 85c8a615a12627c3fb329c85aa14124a0237f5ff is not strictly related to this PR, I think it's an old issue revealed by new tests.

My primary motivation was deep-deep model and predictions explanation support.

Fixes #16


docs/source/libraries/sklearn.rst
```diff
@@ -219,6 +219,17 @@ automatically; to handle HashingVectorizer_ and FeatureHasher_ for
     # and ``ivec`` can be used as a vectorizer for eli5.explain_weights:
     eli5.explain_weights(clf, vec=ivec)
 
+HashingVectorizer_ is also supported inside a FeatureUnion_:
+:func:`eli5.explain_prediction` handles this case automatically, and for
+:func:`eli5.explain_weights` you can use :func:`eli5.sklearn.invert_and_fit``
```
kmike: there is an extra ` at the end of the line

lopuhin: Right, thanks for spotting it, it's so tiny! Fixed in 94d81e9

lopuhin:
@kmike this is ready for review, could you please check it again? I updated the PR description and added a py2 fix.

This feature could be less complicated if InvertableHashingVectorizer would return feature names as strings, not as lists of {'name': , 'sign': } dicts - in this case we would not need code that is currently in ``_invhashing_union_feature_names_scale``. But in this case we would loose the ability to format inverted features nicely, and maybe also work with their signs, etc.

lopuhin:
The main concert I have here is the name of the ``invert_and_fit`` function - it makes sense as a function in ``eli5.sklearn.unhashing``, but maybe for ``eli5.sklearn`` it needs a more descriptive name? We'll be making it public, so it would be harder to change the name.

eli5/sklearn/unhashing.py
```diff
@@ -225,14 +231,89 @@ def _invert_signs(signs):
...
 
 
-def is_invhashing(vec):
-    return isinstance(vec, InvertableHashingVectorizer)
+def invert_and_fit(vec, docs):
+    # type: (Union[FeatureUnion, HashingVectorizer], Any) -> Union[FeatureUnion, InvertableHashingVectorizer]
+    """ Create an InvertableHashingVectorizer from hashing vectorizer vec
+    and fit it on docs. If vec is a FeatureUnion, do it for all
+    hashing vectorizers in the union.
+    Returns an InvertableHashingVectorizer, or a Union, or an unchanged vectorizer.
```
kmike: Union -> FeatureUnion?

kmike:
Yeah, I think it should be named either `invert_hashing_and_fit`, or imported & documented consistently as `sklearn.hashing.invert_and_fit`. The former is probably easier.

lopuhin:
Yes, ``invert_hashing_and_fit`` sounds much better than ``invert_and_fit``, thanks!
Changed the name in b89d0f3

eli5/sklearn/unhashing.py
lopuhin: Right, fixed in b89d0f3. Thanks!

kmike:
Looks good, thanks @lopuhin!

lopuhin:
Thanks for review @kmike !

PR #182, opened 2017-05-03
[DOC] mention eli5.formatters.fields, document "show" argument

lopuhin:
Document ``show`` argument for ``show_prediction`` and ``format_as_text``, and add a mention about constants from ``eli5.formatters.fields``


kmike:
Looks great!

PR #183, opened 2017-05-03
Fix table styles for notebook>=5.0.0

### LINKED ISSUE #178: Horizontal multi-target layout does not look nice with the new notebook
If there are a lot of targets, they might look poorly in the notebook, here is a screenshot from our CRF tutorial
screenshot from 2017-05-02 16-01-24

The scroll is still there, but it's not enough.

I didn't check, but I think it didn't break like that in the previous jupyter notebook version - they clearly changed the CSS quite a bit.

lopuhin:
``table-layout: auto`` fixes the main issue (overrides new default of ``table-layout: fixed``), and adding ``white-space: nowrap;`` for cells with "N more positive/negative" makes the table more tidy.

here is how it looks for full 20 newsgroups dataset:
image

Fixes #178

### DIFF
--- a/eli5/formatters/html.py
+++ b/eli5/formatters/html.py
@@ -71,11 +71,17 @@ def format_as_html(explanation, include_styles=True, force_weights=True,
     return template.render(
         include_styles=include_styles,
         force_weights=force_weights,
-        target_table_styles='border-collapse: collapse; border: none; margin-top: 0em;',
+        target_table_styles=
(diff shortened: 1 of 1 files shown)


kmike:
Does it still look ok in IPython < 5.0.0?

lopuhin:
Yep, I checked, it looks the same in notebook < 5.0.0 apart from nowrap, and nowrap works fine both in old and new versions.

PR #184, opened 2017-05-03
Fix explain_weights for Lasso with one feature

### LINKED ISSUE #180: explain_weights fails for Lasso with fit_intercept=False
```
  File "eli5/sklearn/explain_weights.py", line 377, in explain_linear_regressor_weights
    feature_names = get_feature_names(reg, vec, feature_names=feature_names)
  File "eli5/sklearn/utils.py", line 86, in get_feature_names
    num_features = num_features or get_num_features(clf)
  File "eli5/sklearn/utils.py", line 169, in get_num_features
    return clf.coef_.shape[-1]
IndexError: tuple index out of range
```

lopuhin:
It uses a scalar to store it's weights - this might apply to some other estimators too. Fixes #180

### DIFF
--- a/eli5/sklearn/utils.py
+++ b/eli5/sklearn/utils.py
@@ -124,7 +124,7 @@ def get_default_target_names(estimator, num_targets=None):
     and "y0", "y1", ... if there are multiple targets.
     """
     if num_targets is None:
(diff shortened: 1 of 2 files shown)


kmike:
A good catch!

PR #186, opened 2017-05-03
Make LightGMB with just one job to save memory

### LINKED ISSUE #185: LightGBM sometimes fails to build on Travis
https://travis-ci.org/TeamHG-Memex/eli5/jobs/228232586#L544
```
Building CXX object CMakeFiles/_lightgbm.dir/src/treelearner/gpu_tree_learner.cpp.o
virtual memory exhausted: Cannot allocate memory
```
Maybe making without "-j" would be more reliable? Not sure about any other options - maybe conda has a recent enough lighgbm that is already built?

lopuhin:
Else the build might fail on Travis - checking if it will fix #185

### DIFF
--- a/_ci/install_lightgbm_linux.sh
+++ b/_ci/install_lightgbm_linux.sh
@@ -3,5 +3,5 @@ git clone --recursive https://github.com/Microsoft/LightGBM ; cd LightGBM
 mkdir build ;
 pushd build
 cmake ..
-make -j
+make -j 1
 popd


lopuhin:
yay, the first green build today - I hope that's not a coincidence. On there other hand, there were no failed builds due to lightgbm before, right?

kmike:
Yeah, there were no such failures in past. Maybe the reason is that we're using LightGBM master, so it may start requiring more memory over time to build. We should switch to LightGBM 0.2 as soon as it is released (https://github.com/Microsoft/LightGBM/issues/463). Changes happened too fast though, maybe it is not the cause. Maybe Travis provided more CPU cores for a build, without increasing RAM. Anyways, using 1 process makes sense, thanks for the fix!

PR #188, opened 2017-05-05
explain_prediction with pandas

### LINKED ISSUE #167: Check Pandas and Eli5
Hello kmike. Could you re-check this piece of code? and tell me possible reasons for the error?

lopuhin:
Fixes #167 

Convert dataframe row (Series) to a dataframe with a single element.
XGBoost expects feature names to stay the same, so you can't train it on a dataframe and then pass a numpy array: there will be a feature name mismatch. This is why we need to make it a dataframe, not convert to a 2d numpy array.

Also make determining xgboost feature indices more robust: instead of parsing "f123" strings, look them up in a dictionary of feature names.

TODO:
- [x] check how pandas works with lightning
- [x] check how pandas works with lightgbm
- [x] mention pandas support in the docs and give an example

### DIFF
--- a/eli5/explain.py
+++ b/eli5/explain.py
@@ -97,8 +97,8 @@ def explain_prediction(estimator, doc, **kwargs):
         Example to run estimator on. Estimator makes a prediction for this
         example, and :func:`explain_prediction` tries to show information
         about this prediction. Pass a single element, not a one-element array:
-        if you fitted your estimator on ``xs``, that would be ``xs[i]`` for
-        most containers, and ``xs.iloc[i]`` for ``pandas.DataFrame``.
+        if you fitted your estimator on ``X``, that would be ``X[i]`` for
+        most containers, and ``X.iloc[i]`` for ``pandas.DataFrame``.
 
     top : int or (int, int) tuple, optional
         Number of features to show. When ``top`` is int, ``top`` features with
--- a/eli5/explain.py
+++ b/eli5/explain.py
@@ -96,7 +96,9 @@ def explain_prediction(estimator, doc, **kwargs):
     doc : object
         Example to run estimator on. Estimator makes a prediction for this
         example, and :func:`explain_prediction` tries to show information
-        about this prediction.
+        about this prediction. Pass a single element, not a one-element array:
+        if you fitted your estimator on ``xs``, that would be ``xs[i]`` for
+        most containers, and ``xs.iloc[i]`` for ``pandas.DataFrame``.
 
     top : int or (int, int) tuple, optional
         Number of features to show. When ``top`` is int, ``top`` features with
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -17,6 +17,7 @@ deps=
     pytest-cov
     hypothesis[numpy] !=3.5.1, !=3.5.0
     numpy
+    pandas
```
kmike: It'd be nice to have an environment without Pandas; what about moving pandas dependency to testenv, so that -nodeps environments don't have Pandas?

lopuhin: Right, I wanted to do that but put pandas in the wrong section, thanks for the catch!

kmike:
LightGBM sklearn-compatible wrapper has an explicit pandas support; it'd be great to check how it works. I can do it in another PR, but you're welcome to take a look if you want :)

lopuhin:
@kmike I think this is ready, could you please check it? LightGBM seems to work (the tests pass).

``get_X0`` helper in 381afeb is required to finally convert a dataframe into a numpy array - it worked without it in most cases because ``add_intercept`` did the required conversion implicitly, but it is still required for estimators without intercept, so I use this helper in all cases now instead of previous ``x, = X``

eli5/explain.py
```diff
@@ -96,7 +96,9 @@ def explain_prediction(estimator, doc, **kwargs):
     doc : object
         Example to run estimator on. Estimator makes a prediction for this
         example, and :func:`explain_prediction` tries to show information
-        about this prediction.
+        about this prediction. Pass a single element, not a one-element array:
+        if you fitted your estimator on ``xs``, that would be ``xs[i]`` for
```
kmike: scikit-learn docs use X instead of xs in most places; do you mean something different in ``xs``?

lopuhin: no, let me change it to ``X`` - I mean just all training data here

lopuhin: Fixed in 392e7b1, thanks!

kmike:
Awesome, thanks @lopuhin!

kmike:
Hm, it looks like pandas make test suite much slower (10+ minutes instead of 3 minutes), as for linux it is compiled each time. It also makes tests fail with a timeout - e.g. https://travis-ci.org/TeamHG-Memex/eli5/builds/229209813 and https://travis-ci.org/TeamHG-Memex/eli5/builds/229112950?utm_source=github_status&utm_medium=notification. 

I'm not sure where to get pandas wheels, but I think installing pandas separately first (so that tox uses a wheel) could make it work, as soon as pip produces output during installation. It also seems Travis cache doesn't work for tox environments.

lopuhin:
yeah, 10 minutes is definitely slow :(

> installing pandas separately first (so that tox uses a wheel)

@kmike do you propose building a pandas wheel, putting it somewhere and then installing it from the tox or travis script?

kmike:
@lopuhin yeah; actually, I hope tox will pick it up automatically if it is available in a global pip cache.

PR #189, opened 2017-05-06
Faster pandas build

lopuhin:
Now the tox step is taking about 110 s instead of 718 s. Pandas has wheels available for all python versions we need, it was just not picked up without the wheel package it seems.

So the builds are back to 3-4 minutes with this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ before_install:
     - if [[ "$LIGHTGBM" = 1 ]]; then bash _ci/install_lightgbm_linux.sh; export LIGHTGBM_CHECKOUT=LightGBM; fi
 
 install:
-    - pip install -U pip tox codecov
+    - pip install -U pip wheel tox codecov
 
 script: tox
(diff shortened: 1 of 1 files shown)


kmike:
Weird, I was assuming wheel package is only needed to build wheels, not to use them. Thanks for the fix!

lopuhin:
> Weird, I was assuming wheel package is only needed to build wheels, not to use them. Thanks for the fix!
hmm, actually ``wheel`` is not required - I just checked it with pandas locally, maybe there is something fishy here, let me double-check

lopuhin:
Yeah, the change in this branch is not related to wheel speedup: https://travis-ci.org/TeamHG-Memex/eli5/builds/229393979?utm_source=github_status&utm_medium=notification

Probably the reason for speedup is that pandas wheels were uploaded yesterday (maybe late in the evening?): https://pypi.python.org/pypi/pandas/0.20.1

PR #190, opened 2017-05-06
Faster pandas build, take 2

lopuhin:
This undoes #189 - ``pip`` is able to pick up wheels without ``wheel`` package, and pandas is going to have binary wheels available in the future at the same time as new source release:
https://github.com/pandas-dev/pandas/issues/16260
So we just had some coincidences: adding pandas to our tests, pandas pushing a source release without wheels, and then they fixed them and I pushed an unrelated "fix" :) Sorry for a false fix in #189


kmike:
Haha, a nice catch!

PR #202, opened 2017-05-22
Fix attrs test: see GH-199

### LINKED ISSUE #199: with recent attrs test_base_utils.test_attrs_with_repr fails
```
    def test_attrs_with_repr():
        @attrs
        class WithRepr(object):
            def __init__(self, x):
                self.x = x

            def __repr__(self):
                return 'foo'

>       assert hash(WithRepr(1)) == hash(WithRepr(1))
E       TypeError: unhashable type: 'WithRepr'
```

This is caused by https://github.com/python-attrs/attrs/issues/136; other tests work fine, so it seems `attrs` change doesn't affect eli5. But I can't say it for sure because I don't understand what this test was introduced for in a first place.

lopuhin:
Fixes #199 

This test is meant to check code that makes sure magic methods defined in the class take precedence over attrs-defined ones.

So I think previous ``hash`` was just a mistake - I meant to define some custom ``__hash__`` but forgot about it. I checked that the changed test passes on new attrs as well as on 1.16.3. I also added docstrings to two tricky tests.

### DIFF
--- a/tests/test_base_utils.py
+++ b/tests/test_base_utils.py
@@ -38,6 +38,9 @@ def test_attrs_without_default():
 
 
 def test_attrs_with_repr():
+    """ Check that if __repr__ or __hash__ is defined, attrs does not override
+    them (there is a special logic for it in @attrs).
+    """
 
     @attrs
     class WithRepr(object):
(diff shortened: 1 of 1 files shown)


kmike:
@lopuhin looks good, thanks for the explanation! I don't see docstrings though - are they pushed to github?

lopuhin:
@kmike git ate my docstrings, thanks for the catch! Pushed in 7ea0349

PR #210, opened 2017-05-26
Support xgboost.Booster

### LINKED ISSUE #193: support xgboost.Booster in addition to xgboost sklearn-like interfaces

lopuhin:
Support explain prediction and weights. Fixes #193 

Since the booster does not allow querying parameters that it was trained with (https://github.com/dmlc/xgboost/issues/584), I had two add one required (if booster is passed) argument ``is_regression`` for both ``explain_weights`` and ``explain_prediction``. I also added an optional ``missing`` argument for  ``explain_prediction`` with ``np.nan`` default.

### DIFF
--- a/docs/source/libraries/xgboost.rst
+++ b/docs/source/libraries/xgboost.rst
@@ -13,13 +13,8 @@ xgboost >= 0.6a2.
 .. _XGBRegressor: https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor
 .. _Booster: http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.Booster
 
-:func:`eli5.explain_weights` uses feature importances.
-There is one additional required argument for Booster_ estimator:
-
-* ``is_regression`` True if solving a regression problem
-  ("objective" starts with "reg") and False for a classification problem.
-
-Additional optional arguments for XGBClassifer_, XGBRegressor_ and Booster_:
+:func:`eli5.explain_weights` uses feature importances. Additional
+arguments for XGBClassifer_, XGBRegressor_ and Booster_:
 
 * ``importance_type`` is a way to get feature importance. Possible values are:
 
@@ -43,18 +38,13 @@ of an ensemble. Each node of the tree has an output score, and
 contribution of a feature on the decision path is how much the score changes
 from parent to child.
 
-There is one additional required argument for the Booster_ estimator:
-
-* ``is_regression`` True if solving a regression problem
-  ("objective" starts with "reg") and False for a classification problem.
-
 .. note::
     When explaining Booster_ predictions,
     do not pass an ``xgboost.DMatrix`` object as ``doc``, pass a numpy array
     or a sparse matrix instead (or have ``vec`` return them).
 
 Additional :func:`eli5.explain_prediction` keyword arguments supported
-for XGBClassifer_ and XGBRegressor_:
+for XGBClassifer_, XGBRegressor_ and Booster_:
 
 * ``vec`` is a vectorizer instance used to transform
   raw features to the input of the estimator ``xgb``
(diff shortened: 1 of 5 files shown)


lopuhin:
@kmike now the docs are in place and tests should be green. Could you please review this?

docs/source/libraries/xgboost.rst
```diff
@@ -5,15 +5,21 @@ XGBoost
...
 .. _XGBClassifer: https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier
 .. _XGBRegressor: https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor
+.. _Booster: http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.Booster
 
-:func:`eli5.explain_weights` uses feature importances. Additional
-arguments for XGBClassifer_ and XGBRegressor_:
+:func:`eli5.explain_weights` uses feature importances.
+There is one additional required argument for Booster_ estimator:
+
+* ``is_regression`` True if solving a regression problem
```
kmike: Why is this argument required? It seems for explain_weights it only affects `Explanation.is_regression` flag

lopuhin: Right, maybe we should get rid of it completely? The only place where it actually matters is this (in target_header.html):
```
<b>
    {% if expl.is_regression %}
        {{ target.target }}
    {% else %}
        y={{ target.target }}
    {% endif %}
</b>
```
so if we don't make this distinction and always follow the first branch, then we can get rid of this is_regression flag.

Or did you mean just having is_regression be always False for Booster?

kmike: I meant setting is_regression to None for Booster, but I'm fine with both options. How would explanation look like for regression if we remove the condition?

lopuhin: For explaining xgboost weights, is_regression does not matter at all, because only importances are shown and they don't include the class/target label:
image
For explain_prediction, the difference is this (is_regression=False) 
image

vs. this (is_regression=True)

image

For xgboost, we get proba from actual predictionis, and score from calculation over tree dump - they are equal in case of regression.

Still thinking about this, but I agree that it would be better to remove the is_regression argument or at least make it optional.

lopuhin: @kmike I think I'm happy with solution in dc1b405:
* ``is_regression`` is removed from ``explain_weights`` arguments, cause we don't need it - so it has the same API as for other xgboost estimators
* for ``explain_prediction``, ``is_regression`` is optional. It it is not set and booster has just one target, we assume regression - in this case, proba would not be shown. I think it's less bad to not show proba then to show a wrong proba, and I would not like to compare values of score and proba to decided if proba should be shows. For a multitarget classifier, is_regression is set to False if it was not passed (because there is no multitarget regression in xgboost).

docs/source/libraries/xgboost.rst
```diff
@@ -37,8 +38,13 @@ of an ensemble. Each node of the tree has an output score, and
 contribution of a feature on the decision path is how much the score changes
 from parent to child.
 
+.. note::
+    When explaining Booster_ predictions,
+    do not pass an ``xgboost.DMatrix`` object as ``doc``, pass a numpy array
+    or a sparse matrix instead (or have ``vec`` return them).
```
kmike: Is this limitation just a feature which is not implemented, or is there a deeper reason DMatrix is not supported?

lopuhin: There is a deeper reason, I don't think we'll be able to support it. The reason is that ``DMatrix`` is completely opaque to the python code - you can not convert it to numpy array without dumping to disk (and using some external code to parse it), or read values from it - but we need to be able to read individual values to show the feature values. Due to this, I think it's not a huge drawback - probably people that use DMatrix create it just for training or batch prediction, but also have numpy data that is used for creating DMatrix.

kmike: Got it, thanks!

kmike:
Thanks @lopuhin, the pull request looks great!

lopuhin:
Thanks for review and great suggestions @kmike !

PR #224, opened 2017-06-23
show predicted class in explain_prediction for binary classification

### LINKED ISSUE #223: explain_prediction is confusing for binary classifiers
Currently we're always showing contributions for positive class, not for predicted class:

screen shot 2017-06-23 at 1 48 20 am

Maybe it makes sense to show y=0 for the above example, with positive BIAS and score.

kmike:
Fixes #223.

TODO:

* [x] XGBoost
* [x] tests for XGBoost
* [x] LightGBM
* [x] tests for LightGBM
* [x] sklearn linear models
* [x] lightning linear models
* [x] sklearn tree-based classifiers
* [x] tests for sklearn tree-based classifiers
* [x] add support for `targets` argument in case of binary clasification
* [x] tests for `targets` in case of binary classification
* [x] update docs ~~and notebooks~~ (I'll update notebooks before the release)

I've kept `score` unscaled, so user can see e.g. "y=0: probability 0.708, score -0.883"; feature weights sum to `-1*score` in this case. I'm not sure if that's the way to go or not.

### DIFF
--- a/eli5/sklearn/explain_prediction.py
+++ b/eli5/sklearn/explain_prediction.py
@@ -198,11 +198,15 @@ def explain_prediction_linear_classifier(clf, doc,
             add_weighted_spans(doc, vec, vectorized, target_expl)
             res.targets.append(target_expl)
     else:
-        label_id = 1 if score > 0 else 0
+        if len(display_names) == 1:  # target is passed explicitly
+            label_id, target = display_names[0]
+        else:
+            label_id = 1 if score >= 0 else 0
+            target = display_names[label_id][1]
         scale = -1 if label_id == 0 else 1
 
         target_expl = TargetExplanation(
-            target=display_names[label_id][1],
+            target=target,
             feature_weights=_weights(0, scale=scale),
             score=score,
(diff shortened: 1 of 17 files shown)


kmike:
@lopuhin could you please check it?

eli5/sklearn/explain_prediction.py
```diff
@@ -198,11 +201,18 @@ def explain_prediction_linear_classifier(clf, doc,
...
+            label_id = 1 if score >= 0 else 0
+            target = display_names[label_id][1]
+        scale = -1 if label_id == 0 else 1
+
         target_expl = TargetExplanation(
-            target=display_names[1][1],
-            feature_weights=_weights(0),
+            target=target,
+            feature_weights=_weights(0, scale=scale),
             score=score,
```
lopuhin: Just to clarify - we want to invert the weights and show proba for the predicted class, and show the score unchanged, right? So for negative prediction we should get say "y=0 proba 0.8 score -10", right?

kmike: Yep.

Issue #203, opened 2017-05-22
Add a function to compute "Mean decrease in accuracy" feature importances


kmike:
It seems it shouldn't be model-specific, and it should support metrics other than accuracy.
See https://github.com/scikit-learn/scikit-learn/issues/8898 for more details.

jnothman:
it can also be calculated on a held-out set, such that you measure
importance to generalisation

kmike:
Ah yeah, I was thinking such function should always receive a held-out dataset. Is there a way around it? 

In random forests such datasets are available naturally, so for them MDA feature importances can be calculated as a part of training, without having an explicit held-out dataset. But I'm not sure it can be implemented with unchanged sklearn, xgboost and lightgbm.

jnothman:
well calculating on the training set will be no worse than current
feature_importances except where that is already based on oob estimates

kmike:
Fixed by #227.

PR #227, opened 2017-07-21
[MRG] "score decrease" feature importances

kmike:
An implementation of https://github.com/TeamHG-Memex/eli5/issues/203.


eli5/sklearn/score_decrease.py
```diff
@@ -0,0 +1,230 @@
...
+    """Meta-transformer which exposes feature_importances_ attribute
+    based on average score decrease.
+
+    ScoreDecreaseFeatureImportances instance can be used instead of
+    its wrapped estimator, as it exposes all estimator's common methods like
+    ``predict``.
+
+    There are 3 main modes of operation:
+
+    1. prefit=True (pre-fit estimator is passed). You can call
```
jnothman: In `CalibratedClassifierCV` we went with `cv='prefit'` to express these as mutually exclusive options.

kmike: Thanks @jnothman, a good call. I'll change it.

kmike:
@lopuhin @jnothman I think this is ready; what do you think?

docs/source/blackbox/score_decrease.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Decrease in Score
+=================
+
+eli5 provides a way to compute feature importances for any black-box
+estimator by measuring how score decreases when a feature is not available.
+
+A similar method is described in Breiman, "Random Forests", Machine Learning,
+45(1), 5-32, 2001 (available online at
+https://www.stat.berkeley.edu/%7Ebreiman/randomforest2001.pdf) as
```
lopuhin: I didn't read the whole paper, but tried to find something like "mean decrease in accuracy" and failed - could you please tell the page number @kmike ?

docs/source/blackbox/score_decrease.rst
```diff
@@ -0,0 +1,91 @@
...
+To avoid re-training the estimator we can remove a feature only from the
+test part of the dataset, and compute score without using this
+feature. It doesn't work as-is, because estimators expect feature to be
+present. So instead of removing a feature we can replace it with random
+noise - feature column is still there, but it no longer contains useful
+information. This method works if noise is drawn from the same
+distribution as original feature values (as otherwise estimator may
+fail). The simplest way to get such noise is to shuffle values
+for a feature, i.e. use other example's feature value - this is how
+"decrease in score" feature importances computed.
```
lopuhin: probably "are" is missing: importances **are** computed

docs/source/blackbox/score_decrease.rst
```diff
@@ -0,0 +1,91 @@
...
+For sklearn-compatible estimators eli5 provides
+:class:`~.ScoreDecreaseFeatureImportances` wrapper; if you want to use this
+method for other estimators there is :mod:`eli5.score_decrease` module.
+
+For example, this is how you can check feature importances of
+`sklearn.svm.SVC`_ classifier, which is not supported by eli5 directly
+when a non-linear kernel is used::
+
+    import eli5
+    from eli5.sklearn import ScoreDecreaseFeatureImportances
```
lopuhin: maybe it's better to use ``eli5. blackbox`` for ``ScoreDecreaseFeatureImportances`` import?

docs/source/blackbox/score_decrease.rst
```diff
@@ -0,0 +1,91 @@
...
+
+This method can be useful not only for introspection, but also for
+model selection - one can compute feature importances using
+:class:`~.ScoreDecreaseFeatureImportances`, then drop unimportant features
+using e.g. sklearn's SelectFromModel_. In this case estimator passed to
+:class:`~.ScoreDecreaseFeatureImportances` doesn't have to be fit; feature
+importances can be computed for several train/test splits and then averaged::
+
+    import eli5
+    from eli5.sklearn import ScoreDecreaseFeatureImportances
```
lopuhin: same here, maybe it's better to use ``eli5. blackbox`` for ``ScoreDecreaseFeatureImportances`` import?

lopuhin:
@kmike this looks like an awesome feature, can't wait to try it! Only checked the docs so far, will try to check the rest today, or failing that, on Friday.

docs/source/blackbox/score_decrease.rst
kmike: What do you suggest to put to `eli5.blackbox` module?

lopuhin: Oh sorry, my bad - I thought there is such a module and this was a typo, but now I see that it's only a folder in docs.

docs/source/blackbox/score_decrease.rst
kmike: Yeah, I should be more careful with literature. Method is described on pages 18-19, but it is not called "mean decrease in accuracy" there. It is called "mean decrease in accuracy" here: https://dinsdalelab.sdsu.edu/metag.stats/code/randomforest.html and here: http://alexperrier.github.io/jekyll/update/2015/08/27/feature-importance-random-forests-gini-accuracy.html, and in several papers like this http://ieeexplore.ieee.org/document/7883053/?reload=true. It is also called "Mean Decrease Accuracy" here: http://www.montefiore.ulg.ac.be/%7Eglouppe/pdf/phd-thesis.pdf (see page 125), saying it is proposed in Brieman's papers:

> In addition to MDI, Breiman [2001,2002] also proposed to evaluate the importance of a variable Xj by measuring the Mean Decrease Accuracy (MDA) or, equivalently, by measuring the Mean Increase Error, of the forest when the values of Xj are randomly permuted in the out-of-bag samples. For that reason, this latter measure is also known as the permutation  importance. 

So yeah, I should find a better reference, and at least mention the alternative name (permutation importance) which is also common.

docs/source/blackbox/permutation_importance.rst
```diff
@@ -0,0 +1,115 @@
...
+        PermutationImportance(SVC(), cv=5),
+        threshold=0.05,
+    ).fit(X, y)
+    X_trans = sel.transform(X)
+
+See :class:`~.PermutationImportance` docs for more.
+
+Note that permutation importance should be used for feature selection with
+care. For example, if several features are correlated, and the estimator
+uses them all equally, permutation importance can be low for all of these
```
jnothman: equally, permutation importance may be high for a feature that is only meaningful together with another feature.

jnothman: But I suspect this is somewhat true of many methods that measure feature importance in a model (rather than either associations between multiple features and the target, or univariate relationships without trying to maximise classification performance with all features).

docs/source/blackbox/permutation_importance.rst
```diff
@@ -0,0 +1,115 @@
...
+features: dropping one of the features may not affect the result, as estimator
```
lopuhin: I get the general idea, but if the estimator uses the feature ("using them equally"), then permuting it should still make the score worse?

docs/source/blackbox/permutation_importance.rst
```diff
@@ -0,0 +1,115 @@
...
+    svc = SVC().fit(X_train, y_train)
+    perm = PermutationImportance(svc).fit(X_test, y_test)
+    eli5.show_weights(perm)
+
+.. _sklearn.svm.SVC: http://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html
+
+If you don't have a separate held-out dataset, you can fit
+:class:`~.PermutationImportance` on the same data as used for
+training; this still allows to inspect the model, but doesn't show which
+features are important for generalization.
```
lopuhin: I think a small example of using permutation importance for non-scikit-learn models would be nice - at the moment we link to ``:mod:`eli5.permutation_importance` `` which has two functions and does not give any examples. Not sure which is the best place for this - here or in the module docs.

Maybe an example could look something like this?

```
from eli5.permutation_importance import get_feature_importances

my_model = MyModel()
my_model.fit(X_train, y_train)

def score_func(X, y):
    return my_score(y, my_model.predict(X))

get_feature_importances(score_func, X_valid, y_valid)
```

eli5/permutation_importance.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if columns_to_shuffle is None:
+        columns_to_shuffle = range(X.shape[1])
+
+    if pre_shuffle:
+        X_shuffled = X.copy()
+        rng.shuffle(X_shuffled)
+
+    X_res = X.copy()
+    for columns in columns_to_shuffle:
```
jnothman: Should this be `for column in columns_to_shuffle`?

eli5/permutation_importance.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_feature_importances(score_func, X, y, columns_to_shuffle=None,
+                            random_state=None):
+    """ Return feature importances computed as score decrease when a feature
+    is not available. """
+    score_base = score_func(X, y)
+    Xs = iter_shuffled(X, columns_to_shuffle, random_state=random_state)
+    return np.array([
+        score_base - score_func(X_shuffled, y)
```
jnothman: Do you want this clipped from below at 0?

jnothman: Should we be offering to report the mean of multiple shuffles?

eli5/permutation_importance.py
```diff
@@ -0,0 +1,61 @@
...
+application to random forests. It is known in literature as
+"Mean Decrease Accuracy (MDA)" or "permutation importance".
+"""
+from __future__ import absolute_import
+
+import numpy as np  # type: ignore
+from sklearn.utils import check_random_state  # type: ignore
+
+
+def iter_shuffled(X, columns_to_shuffle=None, pre_shuffle=False,
```
jnothman: I would rename `pre_shuffle` to `shared_permutation` or `shared_order`. `pre_shuffle` describes the implementation; `shared_permutation` aims to describe the semantics.

jnothman: Is there a reason to offer this feature, but not to use it? Would getting the score as the mean of 3 pre-shuffled runs be reasonably efficient and more robust? Is it hard to measure or estimate how much lower the quality of your importance estimates are with `pre_shuffle=True`?

eli5/permutation_importance.py
```diff
@@ -0,0 +1,61 @@
...
+    if there is a lot of columns, or if columns are used multiple times.
+    """
+    rng = check_random_state(random_state)
+
+    if columns_to_shuffle is None:
+        columns_to_shuffle = range(X.shape[1])
+
+    if pre_shuffle:
+        X_shuffled = X.copy()
+        rng.shuffle(X_shuffled)
```
jnothman: If X is big, you will get better memory efficiency by storing the permutation order and permuting each column in the loop.

eli5/sklearn/permutation_importance.py
```diff
@@ -0,0 +1,252 @@
...
+
+        X = check_array(X)
+
+        if self.cv not in (None, "prefit"):
+            self.results_ = self._cv_feature_importances(
+                X, y, groups=groups, **fit_params)
+        else:
+            self.results_ = self._non_cv_feature_importances(X, y)
+
+        self.feature_importances_ = np.std(self.results_, axis=0)
```
jnothman: mean, perhaps? Not enough tests, perhaps? ;)

eli5/permutation_importance.py
kmike: The reason it is called `columns` is that you can pass a list of indices, and shuffle several columns at the same time. It seems this can be useful to catch interactions between features, but I haven't explored how to apply that; it is just semi-supported in this utility function.

eli5/sklearn/permutation_importance.py
kmike: wow! 

I had problems making tests to work with np.std, but there is enough correlation between feature importances and their std deviations (more important features often have larger stddev as well), so I was able made it work by choosing right estimators.

eli5/permutation_importance.py
kmike: In `PermutationImportances` n_iter still works fine if you change this to pre_shuffle, as X is pre-shuffled differently for each iteration. It is a performance optimization which shouldn't affect the result if one just uses all columns once. There were two reasons to implement this feature:

1) pre-shuffling was faster for some reason when a number of columns get larger, probably because of Python overhead;
2) I was thinking it could be useful when we start to shuffle several columns at the same time.

Actually I implemented pre-shuffling first, but then switched to on-demand shuffling, as based on benchmarks it seems faster in common cases, and requires less memory. But in some cases pre-shuffling is faster, so I just kept it.

eli5/permutation_importance.py
kmike: re mean: currently `eli5.permutation_importance` is more like a module with basic building blocks for permutation importance, and eli5.sklearn.PermutationImportance is an end-to-end implementation. Probably I should have written this in a module docstring.

docs/source/blackbox/permutation_importance.rst
kmike: It will make score worse, so importance won't be zero. Let's say we have a feature which makes score worse by 0.3, and 5 highly correlated features which make score worse by 0.5 if you drop them all. If estimator uses them "equally", they should get ~0.1 permutation importance. So if you select a threshold to 0.1, or just tell to keep top-1 feature, all of these correlated features would be removed, while the right thing to do is to remove 4 of them and keep the last. RFE should help in such cases to some extent.

docs/source/blackbox/permutation_importance.rst
kmike: It seems currently `eli5.permutation_importance` is more like a collection of building blocks for permutation importance, not an algorithm on its own. At least you'd likely want to do it starting from multiple random seeds, and likely take a mean of all runs. But yeah, I should probably state it in a module docstring.

eli5/permutation_importance.py
kmike: It seems clipping would result in a biased estimate after when we take a mean - features with a high spread which are clipped at zero may get high non-zero importance.

eli5/permutation_importance.py
kmike: It may need more benchmarks, as one of the reasons to pre-shuffle X was not to permute columns in a loop.

docs/source/blackbox/permutation_importance.rst
kmike: I've updated `eli5.permutation_importance` to be more useful (e.g. n_iter is supported directly now), and also added an example to the docs.

eli5/permutation_importance.py
kmike: The first reason sounds stupid though.

docs/source/blackbox/permutation_importance.rst
lopuhin: @kmike thanks, looks great now!

eli5/permutation_importance.py
jnothman: For fastest permutation of an array, use x.take(order, mode='clip')

jnothman: For fastest permutation of an array, use x.take(order, mode='clip')

kmike: bookmarked, thanks.

kmike:
Draft comparison between permutation importances, absolute coefficient values of linear models and Gini feature importances of decision trees and ensembles: https://gist.github.com/kmike/f978d29a250070a7236c0485b1b1f182. I guess I should check when are scores low, figure out why, and add this to caveats if that's a limitation of permutation importances.

kmike:
Thanks @jnothman and @lopuhin  for the careful review!

I've updated the gist with more metrics and some aggregated tables, but wasn't able to find something we can put to caveats yet. 

It seems most of the questions are addressed, so I'm going to merge it soon. Any objections?

lopuhin:
Looks awesome @kmike 👍

jnothman:
I haven't had a chance to take another look. about to attend relevant ICML
tutorial. if they mention something which makes this more robust, I'll let
you know :)

kmike:
Hey! I'm going to merge it tomorrow and release eli5 0.8, unless there are further comments. Anyways, we can refine this feature later.

PR #229, opened 2017-07-31
transform_feature_names for scalers

kmike:
I can see 3 main ways to show feature names for scalers:

1) display feature names as-is (like it is done in https://github.com/scikit-learn/scikit-learn/pull/6431);
2) show that feature names are scaled/normalized, but hide the details, e.g. `scaled(x1)`;
3) show the complete formula, e.g. `(x1*0.312 - 1.232)` for StandardScaler

In this PR (1) is implemented; at least it is more useful than doing nothing. 

It seems we may want an optional "verbose mode" for feature names, as there are use cases you want the whole formula, and there are use cases you only care about input feature names.

### DIFF
--- a/eli5/sklearn/transform.py
+++ b/eli5/sklearn/transform.py
@@ -38,8 +38,10 @@ def _select_names(est, in_names=None):
 @transform_feature_names.register(MaxAbsScaler)
 @transform_feature_names.register(RobustScaler)
 def _select_scaling(est, in_names=None):
-    return _get_feature_names(est, feature_names=in_names,
-                              num_features=est.scale_.shape[0])
+    if in_names is None:
+        in_names = _get_feature_names(est, feature_names=in_names,
+                                      num_features=est.scale_.shape[0])
+    return [name for name in in_names]
 
 
 # Pipelines
--- a/eli5/sklearn/transform.py
+++ b/eli5/sklearn/transform.py
@@ -8,7 +8,7 @@ from sklearn.linear_model import (  # type: ignore
     RandomizedLogisticRegression,
     RandomizedLasso,
 )
-from sklearn.preprocessing import (
+from sklearn.preprocessing import (  # type: ignore
     MinMaxScaler,
     StandardScaler,
     MaxAbsScaler,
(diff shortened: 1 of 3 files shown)


kmike:
@lopuhin @jnothman what do you think?

eli5/sklearn/transform.py
```diff
@@ -25,6 +31,19 @@ def _select_names(est, in_names=None):
...
+
+@transform_feature_names.register(MinMaxScaler)
+@transform_feature_names.register(StandardScaler)
+@transform_feature_names.register(MaxAbsScaler)
+@transform_feature_names.register(RobustScaler)
+def _select_scaling(est, in_names=None):
+    if in_names is None:
+        in_names = _get_feature_names(est, feature_names=in_names,
+                                      num_features=est.scale_.shape[0])
+    return [name for name in in_names]
```
lopuhin: what does this list comprehension do - is it the same as `list(in_names)`, or you wanted to add something more?

kmike: Yes, I'm converting FeatureNames instance (which comes from `_get_feature_names`) to a list. It is also a left-over from my experiments with more elaborate feature names, when you don't pass names as-is.

eli5/sklearn/transform.py
```diff
@@ -25,6 +31,19 @@ def _select_names(est, in_names=None):
     return [in_names[i] for i in np.flatnonzero(mask)]
 
 
+# Scaling
+
+@transform_feature_names.register(MinMaxScaler)
+@transform_feature_names.register(StandardScaler)
+@transform_feature_names.register(MaxAbsScaler)
+@transform_feature_names.register(RobustScaler)
+def _select_scaling(est, in_names=None):
```
jnothman: I think you mean transform, not select

jnothman:
Outputting feature descriptions as [JsonLogic](http://jsonlogic.com/), perhaps??

kmike:
I like the JsonLogic idea, to output expressions used for computing feature names.

+1 to pull in test changes to make updating #208 easier, but I'm also fine with merging #208 with a few minor changes :)

jnothman:
I'm sure the jsonlogic idea is more attractive to computer scientists than
date scientists though ;)

jnothman:
One advantage of the jsonlogic idea is that sub-expressions are objects
bring reused so it's relatively memory efficient.

kmike:
We would still provide html/text/dataframe/(simplified json?) exports if we use jsonlogic internally, so data scientists should be fine :)

PR #235, opened 2017-08-05
Remove indexes from all exported dataframes

### LINKED ISSUE #234: Remove indexes from dataframes returned by format_as_dataframe
So far from my and @kmike experience we always call ``.reset_index()`` on results of  ``format_as_dataframe``/``explain_*_df``, because usually some data transformation is needed (that's why dataframe format is convenient) and it's usually easier to do that on plain columns.

So the idea is to change the API and always return dataframes without custom indexes.

lopuhin:
Fixes #234. This removes all indexes, and a plain dataframe is used instead of a pivot table for transition features.

Looks like no docs need changing, although it's a breaking change.

I'm using ``df.groupby(['target', 'feature']).agg(lambda x: x)`` in several places in tests, is there a more idiomatic way to say that?

### DIFF
--- a/tests/test_formatters_as_dataframe.py
+++ b/tests/test_formatters_as_dataframe.py
@@ -33,11 +33,13 @@ def test_explain_weights(boston_train):
 
 
 def check_targets_dataframe(df, expl):
-    assert list(df.columns) == ['weight']
+    assert list(df.columns) == ['target', 'feature', 'weight']
+    df_indexed = df.groupby(['target', 'feature']).agg(lambda x: x)
     for target in expl.targets:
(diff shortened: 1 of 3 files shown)


tests/test_formatters_as_dataframe.py
```diff
@@ -130,36 +134,6 @@ def test_bad_list():
         format_as_dataframe([1])
 
 
-def test_targets_with_value():
```
lopuhin: this is already tested above, I forgot to remove this earlier

lopuhin:
@kmike can you please have a look?

kmike:
Thanks @lopuhin!

PR #237, opened 2017-08-10
Freeze lightgbm (2.0.5 segfaults)

lopuhin:
Checking if reverting lightgbm to 2.0.4 will fix the broken build https://travis-ci.org/TeamHG-Memex/eli5/builds/263013487

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,7 +49,7 @@ basepython=python3.5
 deps=
     {[testenv]deps}
     xgboost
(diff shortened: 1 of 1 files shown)


lopuhin:
@kmike what do you think about merging that to fix the build, and then we can check why the new lightgbm segfaults?

kmike:
Sounds good, thanks @lopuhin!

PR #244, opened 2017-11-03
parallel execution for permutation_importance

luoq:


eli5/permutation_importance.py
```diff
@@ -15,6 +15,7 @@
 
 import numpy as np  # type: ignore
 from sklearn.utils import check_random_state  # type: ignore
+from sklearn.externals.joblib import Parallel, delayed
```
kmike: I think you can add `#type: ignore` to this line to make type checks pass

kmike:
Thanks, looks like a great feature!

Tests are failing in Python 2.7; I'm not sure what it takes to make it work. Does it at least work with n_jobs=1?

luoq:
It seems in py2 instance method cannot be pickled in default. https://gist.github.com/luoq/703ec045827caf5a504ff97376db9c3e 
https://stackoverflow.com/questions/1816958/cant-pickle-type-instancemethod-when-using-multiprocessing-pool-map?rq=1

luoq:
It's fixed after python3.5:
https://stackoverflow.com/questions/45901505/python3-6-can-pickle-instance-method-staticmethod-classmethod-but-document-say

thomasfrederikhoeck:
Are you still working on this? Would be very useful.

eyadsibai:
will you drop python2 soon, so this or gets merged?

eyadsibai:
shall we switch this to use https://joblib.readthedocs.io/en/latest/parallel.html instead?

aridf:
Can I ask what happened to this? Is there now a way to use `eli5.sklearn.PermutationImportance()` in parallel?

Issue #259, opened 2018-04-08
failing xgboost tests


derenrich:
So I'm seeing xgboost tests as failing. I started investigating this because I couldn't get the titanic notebook example to run as provided. Then I couldn't get the test suite to run through tox as provided. I had to apply the following patch.

```
diff --git a/docs/source/conf.py b/docs/source/conf.py
index 0d955af..5b7ba71 100644
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -35,22 +35,6 @@ class Mock(MagicMock):
         return Mock()


-MOCK_MODULES = [
-    'lightning',
-    'lightning.impl',
-    'lightning.impl.base',
-    'sklearn_crfsuite',
-    # 'IPython',
-    # 'IPython.display',
-    # 'IPython.lib',
-    # 'typing',
-    'xgboost',
-    'lightgbm',
-    'pandas',
-    'pytest',
-]
-sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
-

 # -- General configuration ------------------------------------------------
diff --git a/tox.ini b/tox.ini
index 6b82311..ebc9039 100644
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,8 @@ deps=
     numpy
     scipy
     scikit-learn
-
+    mock
+    sphinx-rtd-theme

 [testenv]
 whitelist_externals = /bin/bash
```

But once I did that the xgboost tests started failing en masse. I'd guess the underlying API changed under you but looking at xgboost's source history that doesn't seem to be the case so I'm pretty confused.

derenrich:
it's the same exception over and over again.

```
xgb = XGBRegressor(base_score=0.5, booster='gbtree', colsample_bylevel=1,
       col..., reg_lambda=1, scale_pos_weight=1, seed=None,
       silent=True, subsample=1), is_regression = None

    def _check_booster_args(xgb, is_regression=None):
        # type: (Any, bool) -> Tuple[Booster, bool]
        if isinstance(xgb, Booster):
            booster = xgb
        else:
>           booster = xgb.booster()
E           TypeError: 'str' object is not callable

eli5/xgboost.py:225: TypeError
```

haribaskar:
# TypeError: 'str' object is not callable
### prediction_text = explain_prediction(model,X_test.iloc[i])

File "/anaconda/envs/py35/lib/python3.5/site-packages/singledispatch.py", line 210, in wrapper

[...]

Issue #252, opened 2018-01-09
'str' object is not callable during show_prediction (xgboost 0.7)


sudarshan1413:
Hi,

I was trying to execute the exact code mentioned at eli5 documentation over here..

xgboost tutorial : http://eli5.readthedocs.io/en/latest/tutorials/xgboost-titanic.html

but during below step...I am getting an error..
Code:
from eli5 import show_prediction
show_prediction(clf, valid_xs[1], vec=vec, show_feature_values=True)

Error Message:
![image](https://user-images.githubusercontent.com/17721958/34735400-53dec4f8-f595-11e7-9e09-85f9f0832461.png)

Let me know if I am missing anything over here...

My Configuration :- Windows 7, Python 2.7.13 

Any help would be really appreciated !

lopuhin:
Thanks for reporting this @sudarshan1413

This is an issue with recently released xgboost 0.7, eli5 currently does not support it.

sudarshan1413:
Thanks @lopuhin for your quick response...!  Any workaround to solve this.... Will backdated (previous version) xgboost work?  If yes which specific version..

lopuhin:
@sudarshan1413 the last version supported is ``pip install xgboost==0.6a2``, ``pip install 'xgboost<0.7'`` gives the same

sudarshan1413:
Thanks @lopuhin . I will give it a try!

antoinemercier82:
@lopuhin Hi Konstantin, the problem is that it is, at least for me, already quite tricky to install xgboost 0.7 on a Windows machine. I tried to uninstall xgboost 0.7 and install 0.6a2 but with no success. Any chance you would have another workaround?

Thanks

beeva-josealberto:
This works for me in xgboost version 0.7.post3:

`booster = model.get_booster()`
`eli5.show_prediction(estimator = booster, doc = test_set.iloc[0,:])`

Hope it helps!

ianozsvald:
I have a workaround, I have the same problem and I've monkey-patched the function from inside a Jupyter Notebook cell:

```
import eli5

def _check_booster_args(xgb, is_regression=None):
    # type: (Any, bool) -> Tuple[Booster, bool]
    if isinstance(xgb, eli5.xgboost.Booster): # patch (from "xgb, Booster")
        booster = xgb
    else:
        booster = xgb.get_booster() # patch (from "xgb.booster()" where `booster` is now a string)
[...]

PR #268, opened 2018-05-17
Fix build for recent numpy and xgboost

lopuhin:
Summary of changes:
- Fix doctest that started failing due to old numpy
- Fix a test for xgboost 0.7 (#252 #259  )
- Add old xgboost and numpy along with old sklearn to "legacy" tox env
- Fix numpy 1.14 deprecation warnings in ``_numpy_to_python`` and add more tests for it


kmike:
Awesome, thanks @lopuhin! 
Codecov's "83.33% coverage" for a patch looks like an error.

PR #269, opened 2018-05-27
fix leaf_ids type error

haribaskar:
fixed TypeError: object of type 'numpy.int32' has no len()
```
File "C:\Users\username\test.py", line 20, in churn_factor_xgb
    prediction_text = explain_prediction(model,X_test.iloc[i])
  File "C:\ProgramData\Anaconda3\lib\site-packages\singledispatch.py", line 210, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "C:\ProgramData\Anaconda3\lib\site-packages\eli5\xgboost.py", line 196, in explain_prediction_xgboost
    booster, dmatrix, n_targets, feature_names, xgb_feature_names)
  File "C:\ProgramData\Anaconda3\lib\site-packages\eli5\xgboost.py", line 247, in _prediction_feature_weights
    assert len(tree_dumps) == len(leaf_ids)
TypeError: object of type 'numpy.int32' has no len()
```

### DIFF
--- a/eli5/xgboost.py
+++ b/eli5/xgboost.py
@@ -243,7 +243,8 @@ def _prediction_feature_weights(booster, dmatrix, n_targets,
     http://blog.datadive.net/interpreting-random-forests/
     """
     # XGBClassifier does not have pred_leaf argument, so use booster
-    leaf_ids, = booster.predict(dmatrix, pred_leaf=True)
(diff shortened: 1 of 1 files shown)


lopuhin:
Hi, thanks for the pull request, but it looks like the fix is not quite correct, xgboost tests are failing. It would be great if you could also add a test for this issue, or tell how to reproduce it.

PR #281, opened 2018-11-01
Prefer singledispatch from standart library

krkd:
Prefer singledispatch from standard library
fallback on singledispatch package for older python versions

- This pr makes it possible to use eli5 with python 3.7.0
```
******/.env/lib/python3.7/site-packages/singledispatch_helpers.py in get_cache_token()
    157
    158 def get_cache_token():
--> 159     return ABCMeta._abc_invalidation_counter
    160
    161

AttributeError: type object 'ABCMeta' has no attribute '_abc_invalidation_counter'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,5 +58,6 @@ setup(
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
     ],
 )
--- a/setup.py
+++ b/setup.py
@@ -35,13 +35,17 @@ setup(
         'jinja2',
         'numpy >= 1.9.0',
         'scipy',
-        'singledispatch >= 3.4.0.3',
         'six',
         'scikit-learn >= 0.18',
         'typing',
         'graphviz',
         'tabulate>=0.7.7',
     ],
+    extras_require={
+        ":python_version<'3.5.6'": [
+            'singledispatch >= 3.4.0.3',
+        ],
+    },
     classifiers=[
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -35,13 +35,17 @@ def get_long_description():
...
         'scikit-learn >= 0.18',
         'typing',
         'graphviz',
         'tabulate>=0.7.7',
     ],
+    extras_require={
+        ":python_version<'3.5.6'": [
+            'singledispatch >= 3.4.0.3',
+        ],
+    },
```
lopuhin: Would be nice to add `'Programming Language :: Python :: 3.7'` to the list below as well.

krkd:
Thanks for the tip with nodeps. Problem with lates released version 0.5.0 of sklearn-contrib-lightning. Checked master 0.6.dev0 seems ok.

tox.ini
```diff
@@ -94,3 +94,13 @@ deps=
...
+
+
+[testenv:py37-nodeps]
+deps=
+    {[base]deps}
+
+commands=
+    ; without lightning as it is optional
+    pip install -e .
+    bash _ci/runtests_nodeps.sh {posargs: eli5 tests}
```
lopuhin: (refers to review above, forgot to add this comment).
I wonder if ``commands={[testenv:py35-nodeps]commands}`` will work here, since we run the same commands

krkd:
Is it ok to merge this PR or other changes required? 
Thanks

lopuhin:
@krkd sorry for delay, I was totally sure this was already merged. Thank a lot, great job 👍

PR #283, opened 2018-11-17
TST upgrade mypy

kmike:
I've upgraded mypy to latest version. Most changes are related to None handling. Probably we can drop custom `attrs` decorator now, but I haven't done it here.


kmike:
Test failures are addressed here: https://github.com/TeamHG-Memex/eli5/pull/284

PR #291, opened 2019-01-08
Update xgboost titanic tutorial to xgboost==0.81

### LINKED ISSUE #289: Obsolete documentation - tutorial
Hi, 
code in tutorial does not work with latest XGBoost 0.8.1.

Location of the problem:

1. Go here:  https://eli5.readthedocs.io/en/latest/tutorials/xgboost-titanic.html
2. Go to Section 3: Explaining weights - there is code below

```
booster = clf.booster()  # <----- method booster() does not exist anymore
original_feature_names = booster.feature_names
booster.feature_names = vec.get_feature_names()
print(booster.get_dump()[0])
# recover original feature names
booster.feature_names = original_feature_names
```

It throws exception:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-45-2800cf4b424a> in <module>
----> 1 booster = clf.booster()
      2 original_feature_names = booster.feature_names
      3 booster.feature_names = vec.get_feature_names()
      4 print(booster.get_dump()[0])
      5 # recover original feature names

TypeError: 'str' object is not callable
```
which makes sense, because `clf.booster` returns string` 'gbtree'`

lopuhin:
Fixes #289

- use xgboost==0.81
- no warnings any more, remove their filtering
- the issue with xgboost + sparse matrices is no longer happening, so remove ``CSCTransformer``

Also I happened to have a newer pandoc version, which produces slightly different output (adjusted ``sed`` scripts accordingly) - @kmike what do you think, shall we stick to 1.x instead (nbconvert warns that is supports only 1.x)?

### DIFF
--- a/docs/source/_notebooks/xgboost-titanic.rst
+++ b/docs/source/_notebooks/xgboost-titanic.rst
@@ -9,14 +9,14 @@ dataset <https://www.kaggle.com/c/titanic/data>`__, which is small and
 has not too many features, but is still interesting enough.
 
 We are using `XGBoost <https://xgboost.readthedocs.io/en/latest/>`__
-0.6a2 and data downloaded from https://www.kaggle.com/c/titanic/data (it
+0.81 and data downloaded from https://www.kaggle.com/c/titanic/data (it
 is also bundled in the eli5 repo:
 https://github.com/TeamHG-Memex/eli5/blob/master/notebooks/titanic-train.csv).
 
 1. Training data
 ----------------
 
-Let's start by loading the data:
+Let’s start by loading the data:
 
 .. code:: ipython3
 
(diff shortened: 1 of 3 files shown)


kmike:
It seems conversion works with pandoc 2.0, and not only here - https://github.com/jupyter/nbconvert/issues/928 - it looks like there are issues, but they're not that important for us. So I think that's fine not to stick with pandoc 1.x. What do you think about documenting somewhere (maybe in a comment in a script) that the script needs pandoc 2+?

kmike:
A great clean up, thanks @lopuhin!

lopuhin:
Thanks for review @kmike !

> What do you think about documenting somewhere (maybe in a comment in a script) that the script needs pandoc 2+?

I added that to ``docs/README.rst``, and applied new pandoc to all notebooks in #292

PR #292, opened 2019-01-09
Apply pandoc 2.x to all notebooks

lopuhin:
Also adjust ``sed`` scripts: use double back-ticks everywhere, otherwise functions inside code blocks were also replaced.

I checked that ``TF*IDF`` is rendered fine without escaping of ``*`` (and other escape removals).

Followup to #291

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -1,6 +1,6 @@
 Install extra python requirements to be able to build html docs locally::
 
-    pip install sphinx sphinx_rtd_theme -r requirements.txt
+    pip install -r requirements.txt
 
 Install `Pandoc <http://pandoc.org>`_ 2.x for notebook conversion.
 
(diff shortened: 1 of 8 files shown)


docs/README.rst
```diff
@@ -1,5 +1,13 @@
+Install extra python requirements to be able to build html docs locally::
+
+    pip install sphinx sphinx_rtd_theme -r requirements.txt
```
kmike: I wonder if we should add sphinx and sphinx_rtd_theme to docs/requirements.txt

lopuhin: I wondered the same - I thought that ``requirements.txt`` are used by RTD to install extra requirements needed to build the docs. Although just adding sphinx and sphinx_rtd_theme should do no harm if they are already installed - let me do this and if something breaks on RTD we can revert/fix it.

kmike:
Thanks @lopuhin!

PR #295, opened 2019-02-07
Fix the build: fix test for newer numpy

lopuhin:
Exception message is different now:
```
IndexError: index 5 is out of bounds for axis 0 with size 2
```
while it used to be:
```
IndexError: index 5 is out of bounds for axis 1 with size 2
```
the new one makes more sense, as this is a 1d-array.


kmike:
Thanks @lopuhin!

Issue #256, opened 2018-03-13
Conflict with PermutationImportance, DataFrame and XGBoost (with workaround)


ianozsvald:
* Python 3.5
* XGBoost '0.7.post3'
* sklearn '0.19.1'
* Pandas '0.22.0'
* ELI5 '0.8'

I'm working on a regression problem in insurance. XGB outperforms sklearn's tree ensembles. ELI5's show_weights works fine if I use `XGBRegressor` or `RandomForestRegressor`, if I use `PermutationImportance` then `RandomForestRegressor` works but  `XGBRegressor` throws an error. I have a workaround noted below, I'm not sure where the problem lies. Any guidance would be happily received.

Estimator:

```
XGBRegressor(base_score=0.5, booster='gbtree', colsample_bylevel=1,
       colsample_bytree=1, gamma=0, learning_rate=0.1, max_delta_step=0,
       max_depth=2, min_child_weight=1, missing=None, n_estimators=119,
       n_jobs=1, nthread=None, objective='reg:tweedie', random_state=0,
       reg_alpha=0, reg_lambda=1, scale_pos_weight=1, seed=None,
       silent=True, subsample=1, tweedie_variance_power=1.9)

X_test_dbg.shape (1759, 3) 
y_test_dbg.shape (1759,)

type(X_test_dbg) # pandas.core.frame.DataFrame
```

If I train `est` on the `DataFrame` version of the data - training, testing and ELI5's `show_weights` are fine
but `PermutationImportance` throws a feature-name error:
```
perm = PermutationImportance(est).fit(X_test_dbg, y_test_dbg)

ValueError: feature_names mismatch: ['myfeature1', 'myfeature2', 'myfeature3'] ['f0', 'f1', 'f2']
expected myfeature1, myfeature3, myfeature2 in input data
training data did not have the following fields: f1, f0, f2
```

If I change the _type_ of the training data to an `ndarray` then `PermutationImportance` no longer throws an error:
```
X_train_dbg = X_train_dbg.as_matrix()
y_train_dbg = y_train_dbg.as_matrix()
X_test_dbg = X_test_dbg.as_matrix()
y_test_dbg = y_test_dbg.as_matrix()
type(X_train_dbg) # numpy.ndarray for all 4 variables
and then 
est.fit(X_train_dbg, y_train_dbg) 
perm = PermutationImportance(est).fit(X_test_dbg, y_test_dbg)
```

[...]

PR #296, opened 2019-02-12
Support XGBClassifier with pandas DataFrame in PermutationImportance

hofesh:
fix for issue:
https://github.com/TeamHG-Memex/eli5/issues/256

### DIFF
--- a/eli5/sklearn/permutation_importance.py
+++ b/eli5/sklearn/permutation_importance.py
@@ -187,12 +187,12 @@ class PermutationImportance(BaseEstimator, MetaEstimatorMixin):
         """
         self.scorer_ = check_scoring(self.estimator, scoring=self.scoring)
 
+        if pandas_available and isinstance(X, pd.DataFrame):
+            self.scorer_ = self._wrap_scorer(self.scorer_, X.columns)
+
         if self.cv != "prefit" and self.refit:
             self.estimator_ = clone(self.estimator)
             self.estimator_.fit(X, y, **fit_params)
-
-        if pandas_available and isinstance(X, pd.DataFrame):
-            self.scorer_ = self._wrap_scorer(self.scorer_, X.columns)
 
         X = check_array(X)
 
--- a/eli5/sklearn/permutation_importance.py
+++ b/eli5/sklearn/permutation_importance.py
@@ -15,7 +15,10 @@ from sklearn.base import (  # type: ignore
 from sklearn.metrics.scorer import check_scoring  # type: ignore
 
 from eli5.permutation_importance import get_score_importances
+from eli5.sklearn.utils import pandas_available
 
+if pandas_available:
+    import pandas as pd
 
 CAVEATS_CV_NONE = """
 Feature importances are computed on the same data as used for training, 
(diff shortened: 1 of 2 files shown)


eli5/sklearn/permutation_importance.py
```diff
@@ -15,7 +15,10 @@
 from sklearn.metrics.scorer import check_scoring  # type: ignore
 
 from eli5.permutation_importance import get_score_importances
+from eli5.sklearn.utils import pandas_available
 
+if pandas_available:
+    import pandas as pd
```
lopuhin: Looks like the build failure https://travis-ci.org/TeamHG-Memex/eli5/jobs/492103801#L781 is due to missing   ``# type: ignore`` commend at the end here:
```suggestion
    import pandas as pd   # type: ignore
```

hofesh:
@lopuhin added a unit test. not sure about your assumption about having pandas and xgboost when running the tests.

tests/test_sklearn_permutation_importance.py
```diff
@@ -9,11 +9,13 @@
 from sklearn.feature_selection import SelectFromModel
 from sklearn.linear_model import LogisticRegression
 
+from xgboost import XGBClassifier # type: ignore
+import pandas as pd # type: ignore
```
lopuhin: xgboost and pandas are indeed available not in all test environments, to fix this it's possible to put the imports into the test function like this (see https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency):

```
xgboost = pytest.importorskip('xgboost')
pd =  pytest.importorskip('pandas')
```

(also ``# type: ignore`` is not required here, as tests are not type-checked at the moment)

hofesh: fixed

eli5/sklearn/permutation_importance.py
```diff
@@ -151,6 +154,12 @@ def __init__(self, estimator, scoring=None, n_iter=5, random_state=None,
         self.cv = cv
         self.rng_ = check_random_state(random_state)
 
+    def wrap_scorer(self, base_scorer, pd_columns):
```
kmike: would you mjind making this method private (`_wrap_scorer`)? I'm not sure it should be a part of the public API

eli5/sklearn/permutation_importance.py
```diff
@@ -182,6 +191,9 @@ def fit(self, X, y, groups=None, **fit_params):
             self.estimator_ = clone(self.estimator)
             self.estimator_.fit(X, y, **fit_params)
 
+        if pandas_available and isinstance(X, pd.DataFrame):
+            self.scorer_ = self.wrap_scorer(self.scorer_, X.columns)
```
kmike: I think it would be better to move all `self.scorer_` - related code close to each other - maybe move `self.scorer_ = check_scoring(self.estimator, scoring=self.scoring)` line just before this new code, so that they're not separated by `prefit` handling code?

tests/test_sklearn_permutation_importance.py
```diff
@@ -13,7 +13,6 @@
 from eli5.sklearn import PermutationImportance
 from .utils import format_as_all
 
-
```
kmike: according to pep8, function should be separated by two lines, not but one - could you please keep the line here, and add a line before the new test function?

kmike:
I left a few minor comments, but the code looks good to me - thanks @hofesh!

hofesh:
@kmike fixed your comments

kmike:
Thanks @hofesh for the fix!

PR #302, opened 2019-03-27
Fix #298 Handle removal of sklearn randomized l1 models

### LINKED ISSUE #298: Randomized L1 models are removed from sklearn, breaks import
The Randomized Linear models were removed from sklearn but it looks like they are still being imported. That breaks the import.

teabolt:
Fixes #298 
Wrap import and decorators in a try-except block.

### DIFF
--- a/eli5/sklearn/transform.py
+++ b/eli5/sklearn/transform.py
@@ -18,6 +18,7 @@ from eli5.sklearn.utils import get_feature_names as _get_feature_names
 
 # Feature selection:
 
+@transform_feature_names.register(SelectorMixin)
 def _select_names(est, in_names=None):
     mask = est.get_support(indices=False)
     in_names = _get_feature_names(est, feature_names=in_names,
@@ -33,7 +34,6 @@ try:
     _select_names = transform_feature_names.register(RandomizedLogisticRegression)(_select_names)
 except ImportError:     # Removed in scikit-learn 0.21
     pass
-_select_names = transform_feature_names.register(SelectorMixin)(_select_names)
 
 
 # Scaling
--- a/eli5/sklearn/transform.py
+++ b/eli5/sklearn/transform.py
@@ -4,10 +4,7 @@
 import numpy as np  # type: ignore
 from sklearn.pipeline import Pipeline, FeatureUnion  # type: ignore
 from sklearn.feature_selection.base import SelectorMixin  # type: ignore
-from sklearn.linear_model import (  # type: ignore
-    RandomizedLogisticRegression,
-    RandomizedLasso,
-)
+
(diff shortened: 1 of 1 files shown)


eli5/sklearn/transform.py
```diff
@@ -21,15 +18,23 @@
...
+try:
+    from sklearn.linear_model import (  # type: ignore
+        RandomizedLogisticRegression,
+        RandomizedLasso,
+    )
+    _select_names = transform_feature_names.register(RandomizedLasso)(_select_names)
+    _select_names = transform_feature_names.register(RandomizedLogisticRegression)(_select_names)
+except ImportError:     # Removed in scikit-learn 0.21
+    pass
+_select_names = transform_feature_names.register(SelectorMixin)(_select_names)
```
lopuhin: ``SelectorMixin`` could be left as a decorator, right?

eli5/sklearn/transform.py
```diff
@@ -21,15 +18,23 @@
 
 # Feature selection:
 
-@transform_feature_names.register(SelectorMixin)
-@transform_feature_names.register(RandomizedLogisticRegression)
-@transform_feature_names.register(RandomizedLasso)
```
lopuhin: I had an idea for another approach here, but after some thought it looks worse than what is implemented here, still let me put it here:
- put imports at the top, set ``RandomizedLasso `` etc. to ``None`` if they fail to import
- make it possible to write something equivalent to ``@transform_feature_names.register(RandomizedLasso)`` which would do nothing if ``RandomizedLasso`` is ``None``. But it seems that we'd need a wrapper for that, something like ``@register(transform_feature_names, RandomizedLasso)``

Advantage is that the flow would be more natural (imports at the top, decorators above the function). Disadvantage is that it might be a bit confusing that ``RandomizedLasso`` can be ``None``, and also we would need to introduce the wrapper. So to me it seems that it would make sense only if we get a lot of similar conditional registers.

eli5/sklearn/transform.py
teabolt: You are correct. I expanded out and placed `SelectorMixin` last to keep the original order that the decorators are applied in (RandomizedLasso -> RandomizedLogisticRegression -> SelectorMixin). But in this case order does not seem to matter.

eli5/sklearn/transform.py
teabolt: Thanks for the suggestions @lopuhin. Looking at the [recently deprecated](https://scikit-learn.org/stable/modules/classes.html#recently-deprecated) in sklearn there doesn't seem to be too many removals of existing models. But I would certainly prefer your approach if we face more issues like this down the line.

PR #308, opened 2019-04-29
Add scikit-learn-contrib/stability-selection and improve randomized_l1 transformers

### LINKED ISSUE #305: add scikit-learn-contrib/stability-selection support
I think we can add https://github.com/scikit-learn-contrib/stability-selection support, which is mentioned in https://github.com/scikit-learn/scikit-learn/issues/8995 (see also: https://github.com/TeamHG-Memex/eli5/pull/302).

teabolt:
Closes #305.
* Take ```stability_selection.StabilitySelection```(https://github.com/scikit-learn-contrib/stability-selection) as an argument to ```transform_feature_names```. 
* Add stability selection tests.
* Mention stability selection in the docs.
* Register possibly missing dependencies through a wrapper. See @lopuhin's suggestions  https://github.com/TeamHG-Memex/eli5/pull/302#discussion_r269882856. 

Coverage should increase once scikit-learn 0.21 and scikit-learn-contrib/stability-selection are tested against (could modify the tox config for this?).

If more use cases come up for functions like ```register_notnone``` and ```instantiate_notnone```, they could be moved to ```base_utils.py``` or similar.

### DIFF
--- a/docs/source/libraries/sklearn.rst
+++ b/docs/source/libraries/sklearn.rst
@@ -261,6 +261,8 @@ Currently the following transformers are supported out of the box:
 * SelectorMixin-based transformers: SelectPercentile_,
   SelectKBest_, GenericUnivariateSelect_, VarianceThreshold_,
   RFE_, RFECV_, SelectFromModel_, RandomizedLogisticRegression_;
+* stability selection-based transformers: RandomizedLogisticRegression_, 
+  RandomizedLasso_, StabilitySelection_;
 * scalers from sklearn.preprocessing: MinMaxScaler_, StandardScaler_,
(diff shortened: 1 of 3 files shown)


teabolt:
Re: two test parametrizations for RandomizedLasso, one for scikit-learn < 0.19 and another for scikit-learn >= 0.19. It might be possible to reduce these to just one test by passing certain parameters to the RandomizedLasso so that the same features are selected in both scikit-learn versions. See also history of randomized_l1 https://github.com/scikit-learn/scikit-learn/commits/0.20.X/sklearn/linear_model/randomized_l1.py for what might be causing the difference in selected features.

PR #315, opened 2019-06-10
Explain predictions of Keras image classifiers (Grad-CAM)

teabolt:
This PR adds explanations for Keras models that are used to classify images. Specifically we implement [Grad-CAM](https://arxiv.org/pdf/1610.02391.pdf).

For example, the following piece of code:
```
import keras
import numpy as np
import eli5

# load model
xception = keras.applications.xception.Xception(include_top=True, weights='imagenet', classes=1000)

# load image
im = keras.preprocessing.image.load_img('../eli5_examples/motorcycle.jpg', target_size=(299, 299))
doc = keras.preprocessing.image.img_to_array(im)
doc = np.expand_dims(doc, axis=0)
keras.applications.xception.preprocess_input(doc)

# explain
eli5.show_prediction(xception, doc)
```
produces this explanation for the class 'motor scooter':
motorcycle_pr_2

The following features are added:
* Add `keras` package with `explain_prediction_keras()` and `formatters.image` module with `format_as_image()` (requires matplotlib and Pillow).
* Add `.image` attribute to `base.Explanation` and `.heatmap` to `base.TargetExplanation`.
* Make `ipython.show_prediction()` dispatch to an image display function for image explanations.

TODO items before this PR is finalized:
* Resolve reviews
* ~~Coverage~~
* ~~Pass CI~~
* ~~Mypy type annotations~~
* ~~Docs (formatting, tutorial)~~
* ~~Integration and unit tests~~


eli5/ipython.py
```diff
@@ -268,14 +282,51 @@ def show_prediction(estimator, doc, **kwargs):
...
+    Show the heatmap and image overlay in a matplotlib plot 
+    
+    Returns
+    -------
+    figure and axes : tuple[object]
+        Matplotlib objects for the overlay.
+    """
+    # TODO: arguments to this show_prediction_image function 
+    # that control if any labels should be set, i.e. title
+    overlay = format_as_image(expl, **format_kwargs)
```
lopuhin: Another option would be to return a PIL image here, then Jupyter notebook would already be able to display it, right? That would allow to remove matplotlib dependency if we don't need it otherwise. Or is it better to use matplotlib here?

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+import keras.backend as K
+from keras.models import Model
+from keras.layers import Layer, Lambda
+from keras.preprocessing.image import load_img, img_to_array, array_to_img
+
+from eli5.base import Explanation
+from eli5.explain import explain_prediction
+
+
+DESCRIPTION_KERAS = """Grad-CAM visualization for image input; output is images"""
```
lopuhin: > output is images

I'd thought that output is literally an image, and not an explanation object with image and heatmap attributes. I think it would be better to either drop the bit about the output, or rephrase it.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+# note that keras.models.Sequential subclasses keras.models.Model, so we can just register Model
+@explain_prediction.register(Model)
+def explain_prediction_keras(estimator, doc,
+                             target_names=None, # TODO: implement this
```
lopuhin: Sorry if I'm repeating a question you already answered - how do you plan to implement this, where would these target names show up?

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    doc : object
+        An input image as a tensor to ``estimator``, for example ``numpy.ndarray``.
+    target_names : list, optional
+        *`Not Implemented`*.
+
+        Names for classes in the final output layer.
+    targets : list[int], optional
+        Prediction ID's to focus on.
+
+        *`Currently only the first prediction from the list is explained`*.
```
lopuhin: Probably it's better to require a list of length 1 in this case, instead of silently discarding elements after the first.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+        If None, the model is fed the input and its top prediction 
+        is taken as target as the target automatically.
```
lopuhin: Maybe it's an unintended repetition?
```suggestion
        is taken as the target automatically.
```

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    activation_layer = get_activation_layer(estimator, layer)
+    predicted = get_target_prediction(estimator, doc, targets)
+    
+    heatmap = grad_cam(estimator, doc, predicted, activation_layer)
+    # TODO: consider renaming 'heatmap' to 'visualization'/'activations' (the output is not yet a heat map)
+
+    # need to insert a 'channel' axis to get a rank 3 image from rank 2
+    heatmap = np.expand_dims(heatmap, axis=-1)
+    heatmap = array_to_img(heatmap) # -> PIL image
```
lopuhin: I'm curious why do we need to expand axis here and convert it to an image? To me a heatmap as a float numpy array in [0, 1] range would make more sense in the API, although I'm not 100% sure.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    
+    # take the single image from the input 'batch' (rank 4 to rank 3)
+    doc = doc[0]
```
lopuhin: I think it would be nice to clarify that a batch of size one needs to be passed in the API docs, and also to check that the user did pass a batch of size one, raising an error if not. I'm not sure if some models could work with an image without batch dimension - if yes, we might want to check the shape here.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    image = array_to_img(doc) # -> PIL image
+
+    # TODO: return arrays, not images (arrays are more general)
```
lopuhin: FWIW I think it's fine to keep Explanation.image a PIL image.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    heatmap = grad_cam(estimator, doc, predicted, activation_layer)
+    # TODO: consider renaming 'heatmap' to 'visualization'/'activations' (the output is not yet a heat map)
+
+    # need to insert a 'channel' axis to get a rank 3 image from rank 2
+    heatmap = np.expand_dims(heatmap, axis=-1)
+    heatmap = array_to_img(heatmap) # -> PIL image
+    
+    # take the single image from the input 'batch' (rank 4 to rank 3)
+    doc = doc[0]
+    image = array_to_img(doc) # -> PIL image
```
lopuhin: I think it's fine as-is, but in the future we may want to support passing image and neural network input separately - e.g. if the network has some custom input (e.g. 4 channel image, 1 channel image, or something completely custom), then it might be more convenient for the user to pass both input, and also the image which corresponds to this input, which they'll prepare themselves.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # TODO: return arrays, not images (arrays are more general)
+    return Explanation(
+        estimator.name, # might want to replace this with something else, eg: estimator.summary()
+        description=DESCRIPTION_KERAS,
+        error='',
+        method='Vanilla Grad-CAM',
```
lopuhin: I think it's fine to call it just "Grad-CAM"

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+    Raises
+    ------
+    ValueError :
+        If suitable layer can not be found.
+    """
+    # we assume that this is a simple feedforward network
+    # linear search in reverse
+    i = len(estimator.layers)-1
+    while -1 < i and not condition(estimator, i):
```
lopuhin: I think ``i >= 0`` would be a more common way to write this condition

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+    Returns
+    -------
+    is suitable : boolean
+        whether the layer matches what is needed
+    """
+    # TODO: experiment with this, using many models and images, to find what works best
+    # Some ideas: 
+    # check layer type, i.e.: isinstance(l, keras.layers.Conv2D)
+    # check layer name
```
lopuhin: Good idea to experiment. I think that at least this should work at least for all common image classification models.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    # This would be consistent with what is done in https://github.com/ramprs/grad-cam/blob/master/misc/utils.lua
+    # https://github.com/ramprs/grad-cam/blob/master/classification.lua
+    # https://github.com/torch/nn/blob/master/doc/module.md
+    if isinstance(targets, list):
+        # take the first prediction from the list
+        predicted_idx = targets[0]
+        # TODO: use all predictions in the list
+        # TODO: validate list contents
+    elif targets is None:
+        predictions = model.predict(x)
```
lopuhin: Would be great to avoid doing prediction twice (here and when we're explaining the output) if possible, as it's a costly operation

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def grad_cam(estimator, image, prediction_index, activation_layer):
+    """
+    Generate a heatmap using Gradient-weighted Class Activation Mapping (Grad-CAM).
+
+    Credits for implementation
+    * Jacob Gildenblat for "https://github.com/jacobgil/keras-grad-cam".
+    * Author of "https://github.com/PowerOfCreation/keras-grad-cam" for fixes to Jacob's implementation.
+    * Kotikalapudi, Raghavendra and contributors for "https://github.com/raghakot/keras-vis".
```
lopuhin: Great, thanks for providing these links and giving credit 👍

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    
+    Returns
+    -------
+    heatmap : object
+        A numpy.ndarray localization map.
+    """
+    # FIXME: this assumes that we are doing classification
+    # FIXME: we also explicitly assume that we are dealing with images
```
lopuhin: I think both are fine for now, until we add more features.

eli5/keras.py
```diff
@@ -0,0 +1,293 @@
...
+    spatial_shape = activations.shape[:2]
+    lmap = np.zeros(spatial_shape, dtype=np.float32)
+    for i, w in enumerate(weights):
+        # weight * single activation map
+        # add to the entire map (linear combination), NOT pixel by pixel
+        lmap += w * activations[..., i]
+
+    lmap = np.maximum(lmap, 0) # ReLU
+
+    lmap = lmap / np.max(lmap) # -> [0, 1] ndarray
```
lopuhin: If ``np.max(lmap)`` can be zero, would be nice to add some eps, e.g. ``lmap / (eps + np.max(lmap))``, or maybe there is already an existing function for that.

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from PIL import Image
+import matplotlib.cm
+
+
+def format_as_image(expl,
+    interpolation=Image.LANCZOS,
```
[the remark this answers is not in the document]
teabolt: interpolation_comparison (see https://gist.github.com/teabolt/39a38a0587ad0f766933a53a37375d49)

Lanczos looks smoother than bilinear to me. Lanczos is also similar to bicubic.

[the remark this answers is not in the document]
teabolt: One use case for changing the interpolation would be to see the 'true' and not smoothed out activations, as in the 'nearest' and 'box' interpolation examples. 

The drawback is a more complicated API. The user can also just examine the `heatmap` attribute of an `Explanation` instance for the 'true' activations. So I have no problem removing this argument.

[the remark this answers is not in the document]
lopuhin: > One use case for changing the interpolation would be to see the 'true' and not smoothed out activations, as in the 'nearest' and 'box' interpolation examples.

Aha this makes sense, thanks! Let's keep it.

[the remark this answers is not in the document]
lopuhin: > Lanczos looks smoother than bilinear to me. Lanczos is also similar to bicubic.

Thanks for providing examples, I agree that Lanczos looks better here.

eli5/keras.py
teabolt: I want to print, log, or add to the Explanation's `description` the predicted class that Grad-CAM focuses on.

Instead of printing the index of the class, I would print the actual label of the class. For example, for ImageNet with 1000 classes the user could pass a dictionary like [this](https://gist.github.com/yrevar/942d3a0ac09ec9e5eb3a) that we could then use to get the class label.

I think this is similar to what is done in this [tutorial](https://eli5.readthedocs.io/en/latest/tutorials/sklearn-text.html#baseline-model), where the feature names and labels are shown with this code:
```
eli5.show_weights(clf, vec=vec, top=10,
                  target_names=twenty_test.target_names)
```

eli5/ipython.py
teabolt: I think it's better to simply return the PIL image, which you can indeed display with `IPython.display.display()`. If the user wants a fancier plot they can use matplotlib with the returned PIL image.

(matplotlib will still however be a dependency through `formatters/image.py`).

eli5/keras.py
teabolt: I expanded the axis because `array_to_img()` (from keras preprocessing) requires a rank 3 array (the heatmap returned by grad cam is rank 2 / 2D).

eli5/keras.py
lopuhin: Thanks @teabolt , this makes sense 👍

eli5/keras.py
teabolt: I think you are right. If the user accesses `.heatmap`, they can work with the values easily if the contents are a 2D numpy array.

One advantage of converting the heatmap to an image is that the user can readily see how the original heatmap looks. However, it's only a few lines to convert your suggested format into an image, so I think the user can do the conversion themselves.

eli5/keras.py
teabolt: Good catch. We could rely on Keras exceptions. For example, this seems pretty self-explanatory:
Screenshot from 2019-06-13 16-07-35

However, it can be esoteric too:
Screenshot from 2019-06-13 16-08-05

So I will add our own validation function.

eli5/keras.py
teabolt: Just to clarify this idea a bit more. Correct me if I'm wrong please. We want `explain_prediction()` to be able to take two arguments:
1. One is an input to the network, for example a numpy array / tensor. We feed this input directly into the network and do calculations with it.
2. The second argument is a "displayable" image that corresponds to the input argument, for example a PIL Image object. We don't do anything with this image in `explain_prediction()`, but we can use it in `format_as_image()` to overlay the heatmap over it?

This fixes the problem when the actual input tensor is complex and can not be simply converted to a PIL image by the above two lines of code.

lopuhin: @teabolt yes, this is exactly what I had in mind 👍

lopuhin: @teabolt yes, this is exactly what I had in mind 👍

docs/source/tutorials/keras-image-classifiers.rst
```diff
@@ -0,0 +1,12 @@
...
+
+.. note::
+
+    This tutorial is intended to be run in an IPython notebook.
+    It is also available as a notebook file here_.
+
+.. _here: https://github.com/TeamHG-Memex/eli5/blob/master/notebooks/keras-image-classifiers.ipynb
+
+.. 
+    note that we need to rewrite the image paths when the notebook is converted to rst
```
lopuhin: Can this be automated? We already have some similar fixes in the script that converts notebooks.
Also this note probably shouldn't be in user-visible docs?

eli5/base.py
```diff
@@ -25,6 +25,8 @@ def __init__(self,
                  decision_tree=None,  # type: Optional[TreeInfo]
                  highlight_spaces=None,  # type: Optional[bool]
                  transition_features=None,  # type: Optional[TransitionFeatureWeights]
+                 image=None, # type: object
```
lopuhin: If this can be anything, it's better to use ``Any`` type, but for image and heatmap it seems it's possible to be more specific? E.g. ``Optional[np.ndarray]``?

eli5/formatters/image.py
```diff
@@ -0,0 +1,326 @@
...
+from PIL import Image # type: ignore
+import matplotlib.cm # type: ignore
+
+from eli5.base import Explanation
+
+
+# FIXME: in autodoc the interpolation and colormap defaults show
+# something like: 
+# interpolation=<Mock id='139885759460352'>, colormap=<Mock id='139885771653072'>
+# ID's instead of actual names.
```
lopuhin: Maybe we can install matplotlib when building docs to fix this, and not mock it?

eli5/formatters/image.py
```diff
@@ -0,0 +1,326 @@
...
+        Either a colormap from matplotlib.cm, 
+        or a callable that takes a rank 2 array and 
+        returns the colored heatmap as a [0, 1] RGBA numpy array.
+
+        Example colormaps from matplotlib.cm
+            * ``viridis``
+            * ``jet``
+            * ``binary``
+        
+        See also https://matplotlib.org/3.1.0/gallery/color/colormap_reference.html.
```
lopuhin: It's possible to link to the latest version:
```suggestion
        See also https://matplotlib.org/gallery/color/colormap_reference.html.
```

eli5/formatters/image.py
```diff
@@ -0,0 +1,326 @@
...
+
+    Returns
+    -------
+    resized_image : PIL.Image.Image
+        A resized PIL image.
+    """
+    # PIL seems to have a much nicer API for resizing than scipy (scipy.ndimage)
+    # Also, scipy seems to have some interpolation problems: 
+    # https://github.com/scipy/scipy/issues/8210
+    spatial_dimensions = (image.height, image.width)
```
lopuhin: I think PIL uses (width, height) order

eli5/ipython.py
```diff
@@ -268,14 +291,30 @@ def show_prediction(estimator, doc, **kwargs):
...
     format_kwargs, explain_kwargs = _split_kwargs(kwargs)
     expl = explain_prediction(estimator, doc, **explain_kwargs)
-    html = format_as_html(expl, **format_kwargs)
-    return HTML(html)
+    if expl.image is not None and expl.heatmap is not None:
+        # dispatch to image display implementation
+        if display_prediction_image is not None:
+            return display_prediction_image(expl, **format_kwargs)
+        else:
+            print('Missing dependencies. Can not display as image.')
```
lopuhin: I think it's better to ``warn`` and not print, and also to specify which dependencies are missing.

eli5/keras.py
```diff
@@ -0,0 +1,465 @@
...
+
+
+    :raises ValueError: if ``doc`` shape does not match.
+    """
+    input_sh = estimator.input_shape
+    doc_sh = doc.shape
+    if len(input_sh) == 4:
+        # rank 4 with (batch, ...) shape
+        # check that we have only one image (batch size 1)
+        single_batch = (1, input_sh[1], input_sh[2], input_sh[3])
```
lopuhin: Might be a bit cleaner with ``(1,) + input_sh[1:]``

tests/test_keras.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.fixture(scope='module')
+def simple_seq():
+    """A simple sequential model for images."""
+    model = Sequential([
+        Activation('linear', input_shape=(32, 32, 1)), # index 0, input
+        Conv2D(10, (3, 3)),                            # index 1, conv
+        Conv2D(20, (3, 3)),                            # index 2, conv2
+        gap_layer,                                     # index 3, gap
+    ])
+    # model.summary() # TODO: print model summary if fail
```
lopuhin: I think you can do ``print(model.summary())`` and this would be shown only if the test fails, because pytest does not show test stdout when they pass (not 100% sure it applies to fixtures though, if not you can print in the test).

docs/source/tutorials/keras-image-classifiers.rst
teabolt: > Can this be automated? We already have some similar fixes in the script that converts notebooks.

Thanks for bringing this up. Completely missed that there is a `update-notebooks.sh` script. Will add commands to that.



>  Also this note probably shouldn't be in user-visible docs?

This is a Sphinx "comment" so it wouldn't be visible to the user. But since we have a script there is no need for it.

eli5/base.py
teabolt: Good catch. I will annotate heatmap with `Optional[np.ndarray]` and keep image as `Any` (don't want to import `PIL` as it is optional).

In the future we can start using the solution presented at https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles, where we place imports needed only for annotations under a `if TYPE_CHECKING:` statement.

eli5/formatters/image.py
teabolt: Indeed installing `matplotlib` (and `PIL`) removed the `<Mock...` artefacts. 

Unfortunately autodoc actually evaluates the expressions so I get something like `interpolation=1, colormap=<matplotlib.colors.ListedColormap object>`. It is possible to override the function signature doc (https://stackoverflow.com/a/12087750/11555448), which is a solution that I would prefer.

eli5/formatters/image.py
teabolt: You are right.

Worth mentioning that, unlike Pillow, Keras uses (height, width) order.

tests/test_keras.py
teabolt: Perfect. Works even if I print inside a fixture.

teabolt: Would be nice to somehow display the images too. Though I'm not sure how can we automate this through Travis.

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/80ff9ec716f86829d22918440d0a97913813da37

docs/source/tutorials/keras-image-classifiers.rst
lopuhin: > This is a Sphinx "comment" so it wouldn't be visible to the user. But since we have a script there is no need for it.
 TIL about Sphinx comments, thanks!

eli5/base.py
lopuhin: I see, this makes sense, thank you!

eli5/formatters/image.py
lopuhin: Thanks for the fix! Yeah most DL frameworks use this order, because of how images are stored in arrays (with height dimension first), I think this confusing.

tests/test_keras.py
lopuhin: Nice! Yeah not sure how we can show images on travis. But hopefully we should not have any cases where a test fails on travis but not locally.

eli5/formatters/image.py
lopuhin: > It is possible to override the function signature doc

That looks great, thanks!

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from PIL import Image
+import matplotlib.cm
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/2640623a9d5bab0d169264f1d715a29fefb120f3

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def format_as_image(expl,
+    interpolation=Image.LANCZOS,
+    colormap=matplotlib.cm.magma,
+    alpha_limit=165.75,
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/6eb8ce3693528c4332a0bbd3b4482c31f351aa2d

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    if isinstance(starting_array, np.ndarray):
+        alpha = starting_array
+    else:
+        # take the alpha channel as is
+        alpha = image_array[:,:,3]
+    if alpha_limit is not None:
+        alpha = np.minimum(alpha, alpha_limit)
+    image_array[:,:,3] = alpha
+    return image_array
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/afa5f53e2af44846e18c96321f9005befcb339a5

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
...
+    
+    Returns
+    -------
+    pil_image : object
+        A PIL image object.
+    """
+    if isinstance(img, np.ndarray):
+        img = Image.fromarray(img) # ndarray -> PIL image
+    if isinstance(img, Image.Image):
+        if img.mode == 'RGB':
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/c1d3f7efa24a5448eca773bfd51a7c71d6c374c5

eli5/formatters/image.py
```diff
@@ -0,0 +1,157 @@
...
+            img = img.convert(mode='RGBA') # RGB image -> RGBA image
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/c1d3f7efa24a5448eca773bfd51a7c71d6c374c5

eli5/ipython.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/c357cbd4fcb245f5260436af8b434c223fb1ffca

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/2cf16366cf207ebd3d625ef9ce0aa97a40983762

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/00ae8538385151d1fb42d82ce0da143bc9055f99

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/20d3635c8468f1702b619264f1296d4be1dbe283

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/2ce3ac396cbe49202eb73b18639f2aaa815baa29

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/9744451a5d3ddb137c5631fde450f2197656d442

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/2821b25830a7dbca1dccbfa64f8e92a166908b6a

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/103d9f03a09e7e8328e78f9c1fe8ff2935db0f42

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/f82955188f63a649ad29845763af17079237b11d

eli5/keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/9151d4ca5c30656fd8305185935769025ff67575

docs/source/_notebooks/keras-image-classifiers.rst
```diff
@@ -0,0 +1,435 @@
...
+    import keras
+    from keras.applications import mobilenet_v2
+    
+    import eli5
+    
+    # let's instantiate our model
+    model = mobilenet_v2.MobileNetV2(include_top=True, weights='imagenet', classes=1000)
+    
+    # and load our sample image
+    image = 'imagenet-samples/cat_dog.jpg'  # TODO: replace this image with something more interesting
```
lopuhin: > TODO: replace this image with something more interesting

I think this image is perfectly fine

docs/source/_notebooks/keras-image-classifiers.rst
```diff
@@ -0,0 +1,435 @@
...
+    
+    dims = model.input_shape[1:3] # 224x224
+    doc = eli5.keras.image_from_path(image, image_shape=dims)
```
lopuhin: I think this is not the best way to load an image for the purpose of tutorial, because the user already knows how to load images, and might have images already loaded, but it's not clear from this what is ``doc`` and how to convert an already loaded image to this. I would suggest  to load an image in some way which would be recognized by the user (e.g. Image.open or keras.preprocessing.image.load_img), show the type and dimensions of resulting image, then apply any post-processing if needed, and again show result.

docs/source/_notebooks/keras-image-classifiers.rst
```diff
@@ -0,0 +1,435 @@
...
+
+2. Choosing the target class (target prediction)
+------------------------------------------------
+
+We can make the model classify other objects and check where the
+classifier looks to find those objects.
+
+.. code:: ipython3
+
+    cat_idx = 282 # ImageNet ID for "tiger_cat" class, because we have a cat in the picture
```
lopuhin: Maybe it would be helpful to show indices and probabilities of top 5 or 10 predictions from the network, I imagine this is the way you obtained these indices, it's not some random cat class but the highest scoring one?

docs/source/_notebooks/keras-image-classifiers.rst
```diff
@@ -0,0 +1,435 @@
...
+
+    NASNet
+    Taking top prediction: 243
+
+
+
+.. image:: ../_notebooks/keras-image-classifiers_files/keras-image-classifiers_33_3.png
+
+
+Wow ``show_prediction()`` is so robust!
```
lopuhin: That's a great tutorial @teabolt , very engaging and covers the topic nicely 👍

docs/source/libraries/keras.rst
```diff
@@ -0,0 +1,40 @@
...
+Keras
+=====
+
+Keras_ is "a high-level neural networks API, written in Python and capable of running on top of TensorFlow, CNTK, or Theano". 
+
+Currently ELI5 supports :func:`eli5.explain_prediction` for Keras image classifiers.
+
+.. _Keras: https://keras.io/
+
+:func:`eli5.explain_prediction` explains image classifications through `Grad-CAM <https://arxiv.org/pdf/1610.02391.pdf>`_. The :class:`eli5.base.Explanation` object returned has a ``.image`` attribute that represents the  image that is inputted into the model, and a ``.heatmap`` attribute that is a grayscale "localization map" roughly indicating regions of importance in the image for the predicted class.
```
lopuhin: Usually we prefer to break long lines, unless it's necessary for formatting. Also probably a small typo her:
```suggestion
:func:`eli5.explain_prediction` explains image classifications through `Grad-CAM <https://arxiv.org/pdf/1610.02391.pdf>`_. The :class:`eli5.base.Explanation` object returned has an ``.image`` attribute that represents the  image that is inputted into the model, and a ``.heatmap`` attribute that is a grayscale "localization map" roughly indicating regions of importance in the image for the predicted class.
```

docs/source/libraries/keras.rst
```diff
@@ -0,0 +1,40 @@
...
+
+Important arguments for ``Model`` and ``Sequential``:
+
+* ``doc`` is an image as a tensor that can be inputted to the model.
```
lopuhin: Do we have any more specific requirements? Must it have batch dimension? Specific layout (e.g. BHWC)? Do we support anything besides numpy arrays? Docstring below in ``explain_prediction_keras`` is much more specific, I think it makes sense to be as specific here.

eli5/formatters/image.py
```diff
@@ -0,0 +1,324 @@
...
+    update_alpha(heatmap, starting_array=heat_values, alpha_limit=alpha_limit)
+
+    heatmap = heatmap_to_rgba(heatmap)
+
+    heatmap = resize_over(heatmap, image, interpolation=interpolation)
+    overlay = overlay_heatmap(heatmap, image)
+    return overlay
+
+
+def heatmap_to_grayscale(heatmap):
```
lopuhin: Will this function and other public function below appear in the docs? If yes, I think it would be better to  mark all functions which are not in the public API as private (starting with an underscore, or maybe hide them in another way?). For this and similar functions, I think we don't expect the user to call them? Making them private makes API and docs smaller, and allows us to refactor code in the future more freely.

eli5/ipython_image.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from typing import Any
+
+from PIL import Image # type: ignore
+
+from .base import Explanation
+from .formatters import format_as_image
+
+
+def display_prediction_image(expl, **format_kwargs):
```
lopuhin: This function does nothing except for calling ``format_as_image``, do we need it?

eli5/keras.py
```diff
@@ -0,0 +1,465 @@
...
+        error='',
+        method='Grad-CAM',
+        is_regression=False, # TODO: classification vs regression model
+        highlight_spaces=None, # might be relevant later when explaining text models
+        image=image, # PIL image
+        heatmap=heatmap, # 2D [0, 1] numpy array
+    )
+
+
+def validate_doc(estimator, doc):
```
lopuhin: Same question regarding visibility in API docs - if these functions are visible, let's hide them.

eli5/keras.py
```diff
@@ -0,0 +1,465 @@
...
+
+    Credits
+        * Jacob Gildenblat for "https://github.com/jacobgil/keras-grad-cam".
+        * Author of "https://github.com/PowerOfCreation/keras-grad-cam" for fixes to Jacob's implementation.
+        * Kotikalapudi, Raghavendra and contributors for "https://github.com/raghakot/keras-vis".
+    """
+    # Get required terms
+    weights, activations, grads, predicted_idx, score = grad_cam_backend(estimator, doc, targets, activation_layer)
+
+    # TODO: might want to replace np operations with keras backend operations
```
lopuhin: I think it's fine to have them as numpy - when we add pytorch support in the future, this code would likely be re-used.

eli5/keras.py
```diff
@@ -0,0 +1,465 @@
...
+    # output of target layer, i.e. activation maps of a convolutional layer
+    activation_output = activation_layer.output 
+
+    grads = K.gradients(score, [activation_output])
+    # FIXME: this might have issues
+    # See https://github.com/jacobgil/keras-grad-cam/issues/17
+    # a fix is the following piece of code:
+    # grads = [grad if grad is not None else K.zeros_like(var) 
+    #         for (var, grad) in zip(xs, grads)]
+    grads = grads[0]
```
lopuhin: When can we have more gradients? Do we want to use them? Maybe silently discarding them is ok, but it would be good to add an explanation here.

tests/test_keras.py
lopuhin: Can this TODO be removed now?

eli5/keras.py
teabolt: Our current `grads` definition (`grads = K.gradients(score, [activation_output])`) has only one variable for "ys" (`activation_output`), so there will be only one tensor in the `grads` singleton list. (`K.function` and other operations require tensors).

(Added comment about this)

lopuhin: Thanks @teabolt 👍 In this case, I suggest adding an assert or writing something like ``grads, = grads``, so that if there are more gradients than we expect, we'd know about this.

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: Indeed using functions recognisable by the user is clearer. Should I remove the `eli5.keras.image_from_path()` function from the API? May be too specialised.

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: It would be cool to include this with the `Explanation` object as well. Perhaps a new attribute like `.output` or `.predictions` would be required?

eli5/formatters/image.py
teabolt: Using an underscore works well I think, since the docs don't show such functions automatically.

teabolt: I think `heatmap_to_grayscale()` and `heatmap_to_rgba()` are okay as public functions for convenience? The rest can be hidden except for `format_as_image()`.

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/4d4556e88b07bb89a101495733fae05ef62df3ed

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: Also should I remove the `eli5.keras.image_from_path()` function from the API? May be too specialised.

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/34e3cee6bb51dabfe867dcaf67888bcd9f0ad30f

eli5/ipython_image.py
teabolt: The original idea was to use separate files because the image display method has extra dependencies, and it is not related to the HTML display method. Though for this PR I think it's fine keeping everything in one file.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/998d135d7d050fd23c0a10e065f840555d1d6cee

tests/test_keras.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/f2260b29c362ae680404f8de784344de0df268e1

eli5/ipython.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/d2ed75be5e2ecf19b62b20da0dfa2fcc93079c43

eli5/keras.py
teabolt: Perhaps `grad_cam()` and `grad_cam_backend()` functions should be kept public? This way the user can check the values of the gradients, weights, activations, etc.

kmike: What do you think about making eli5.keras a package, and then have eli5.keras.grad_cam module, + maybe eli5.keras.utils?

kmike: if grad_cam is not keras-specific, then it can be eli5.gradcam as well.

teabolt: Sounds like a good way to partition things.

Though I am not sure whether I should have `eli5.keras.gradcam` or `eli5.gradcam`. I could have both, placing code that could be re-used in `eli5.gradcam` (see @lopuhin's comment about pytorch support just down below), and Keras-specific code in the other.

lopuhin: I think it's fine if we eventually have both ``eli5.gradcam`` and ``eli5.keras.gradcam``, but for now I think it's fine to just have `eli5.keras.gradcam`` and factor out common code when we add pytorch support.

lopuhin: > Perhaps grad_cam() and grad_cam_backend() functions should be kept public? This way the user can check the values of the gradients, weights, activations, etc.

Sure, if you think they can be useful let's keep them public.

eli5/formatters/image.py
lopuhin: Yeah, it's fine to keep such helpers public, and I like `expand_heatmap` name

docs/source/_notebooks/keras-image-classifiers.rst
lopuhin: Now that I think about this again, I wonder if we could use ``TargetExplanation`` for that?

docs/source/_notebooks/keras-image-classifiers.rst
lopuhin: > Also should I remove the eli5.keras.image_from_path() function from the API? May be too specialised.

Yes, I'd rather remove it from the API

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: `TargetExplanation` has a required `feature_weights` argument. I am not sure if the argument would be applicable in this case. I could make it an optional parameter?

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/a923864ae355b5597ec398a33b9cd9f43f2c4250

eli5/formatters/image.py
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/88417feb979ac722d1fbbc9ffb8a5baf4fec365c

eli5/keras.py
teabolt: Great solution @lopuhin thanks.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/684465b048f1a9f7169d02a241a842246b8cd8c7

docs/source/libraries/keras.rst
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/64a31a190f50d9446ae029fa4e24f88983de9f22

docs/source/_notebooks/keras-image-classifiers.rst
```diff
@@ -33,56 +35,210 @@ To start out, lets load our image classifier and data.
...
+
+
+.. code:: ipython3
+
+    # one last thing
+    # `keras.applications` models come with their own input preprocessing function
+    # for best results, apply that as well
+    
+    # mobilenetv2-specific preprocessing
+    # (this operation is in-place)
```
lopuhin: If this is in-place, I think it's better to remove assignment and have only ``mobilenet_v2.preprocess_input(doc)``

lopuhin:
@teabolt regarding ``TargetExplanation``, I looked again and now it looks like a better place for ``.heatmap`` attribute (not sure about image). In some sense, ``.heatmap`` is similar to ``.feature_weights`` attribute, and also explanation present in heatmap is an explanation for one target - so if we add support for explaining multiple targets separately, it would make sense to put them into separate ``TargetExplanation`` objects. And the same goes to text: we'd put text explanation into ``.weighted_spans`` attribute which is already present on ``TargetExplanation``. Regarding ``.image`` attribute, it's less clear-cut, I think it makes sense to put it where it's most convenient. And you are right that we'd need to make ``.feature_weights`` not required if we do this way. What do you think?

teabolt:
@lopuhin making `.heatmap` an optional attribute of `TargetExplanation` sounds great. I think we can keep `.image` as an attribute of `Explanation`, since different `TargetExplanation`'s will still explain the same image?

For `TargetExplanation` I think we can make all attributes optional except for `target`. We can also make `target` accept not only `str` but `int` (class ID) as well?

`TargetExplanation` also has `proba` and `score` attributes. I think for this PR I can just put the predicted value for `target` in the `proba` attribute, and leave out `score`?

lopuhin:
> making .heatmap an optional attribute of TargetExplanation sounds great. I think we can keep .image as an attribute of Explanation, since different TargetExplanation's will still explain the same image?

I think that's a great option, if this does not pose any difficulties in code.

> For TargetExplanation I think we can make all attributes optional except for target. We can also make target accept not only str but int (class ID) as well?

Sounds good to me

> TargetExplanation also has proba and score attributes. I think for this PR I can just put the predicted value for target in the proba attribute, and leave out score?

Yes, that's an option. Ideally, if the passed network does not have softmax, then it's output should be interpreted as a score, and if it has, as a proba. Not sure how to best achieve this - can we check if the last layer does not have any activations?

teabolt:
> can we check if the last layer does not have any activations?

Indeed we can check the last layer's `.activation` attribute. If the attribute value is equal to `keras.activations.linear`, the layer has no activation function.

docs/source/_notebooks/keras-image-classifiers.rst
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/5d805945fc7a2c8287a86c63e422a0f0b7efba18

teabolt:
Making `feature_weights` an optional parameter of `TargetExplanation` posed significant difficulties (a lot of existing code requires this argument). As a replacement I created an `empty_feature_weights` instance for "mocking" `feature_weights`.

The full changes are here: https://github.com/TeamHG-Memex/eli5/pull/315/commits/60fd88125625b3c2a0bfaea904c3693343512846

lopuhin:
Thanks @teabolt , changes look good overall, I have one question regarding

> Making feature_weights an optional parameter of TargetExplanation posed significant difficulties (a lot of existing code requires this argument).

Which kind of difficulties did you meet? Do you mean typechecking failures? If yes, we can discuss several ways to solve them, maybe they would be better than adding ``empty_feature_weights`` - I don't particularly like it as it seems more of a hack and can also be confusing to the user, because they won't see the definition of ``empty_feature_weights`` and would see just some weird looking weights.

teabolt:
> Do you mean typechecking failures? If yes, we can discuss several ways to solve them, maybe they would be better than adding `empty_feature_weights`

Yes the main problem is mypy type checking. One solution I tried was to wrap every use of `feature_weights` in an `if feature_weights is not None` block, but we would need many such blocks. I agree that `empty_feature_weights` is a hack. What other solutions do you have in mind?

teabolt:
> It's not only softmax that can give probabilities, sigmoid also can

Ah I was wondering about that. Sigmoid gives values between 0 and 1 so they can be interpreted as probabilities, though the values don't necessarily sum up to 1 (http://dataaspirant.com/2017/03/07/difference-between-softmax-function-and-sigmoid-function/).

Should I include `keras.activations.hard_sigmoid(x)` as well? (https://keras.io/activations/)

Another concern is models with custom activation functions that output "probabilities".

lopuhin:
> Another concern is models with custom activation functions that output "probabilities".

Yeah, another approach could be to declare except for no activation as outputting probabilities, but then there are also log_softmax layers that output log-probabilities... Maybe instead we could give up and declare everything scores for now, not sure.

teabolt:
> Maybe instead we could give up and declare everything scores for now, not sure.

I could make a mention in the docs that `.proba` is set if final layer is `softmax` or `sigmoid`?

lopuhin:
@teabolt yes, looks good to me

lopuhin:
> One solution I tried was to wrap every use of feature_weights in an if feature_weights is not None block, but we would need many such blocks. I agree that empty_feature_weights is a hack. What other solutions do you have in mind?

Another solution is instead of an if block, do ``assert x.feature_weights is not None``. And maybe it would be possible to do this in just a few places which are then dispatched into other functions (but not sure if it's possible to have just a few places). Another approach, again not sure if it'll work, will be to have ``FeatureWeights([], [])`` as empty feature weights - it feels a bit better to me :)

> but we would need many such blocks

how many would that be? Adding 5-10 asserts does not sound too bad.

teabolt:
> instead of an if block, do `assert x.feature_weights is not None`

 I managed to apply this fix, great idea. Note that where `feature_weights` is used in a list/generator comprehension, I used a `if x.feature_weights is not None` condition. 

This leads to two outcomes: 1. We fail if *any* targets have `feature_weights` set to None. 2. We *skip* whatever we do with `feature_weights` for targets that have it as None.

Feel free to review https://github.com/TeamHG-Memex/eli5/pull/315/commits/017bf3e6d2ed367cd51c96397716401b7b6f953b for any changes.

> how many would that be? Adding 5-10 asserts does not sound too bad.

I have 6 changes across 5 files. Indeed it's not too bad especially with mypy.

lopuhin:
> Feel free to review 017bf3e for any improvements.

Looks perfect, thank you! 👍

eli5/keras/explain_prediction.py
```diff
@@ -0,0 +1,253 @@
...
+# note that keras.models.Sequential subclasses keras.models.Model
+@explain_prediction.register(Model)
+def explain_prediction_keras(estimator, # type: Model
+                             doc, # type: np.ndarray
+                             target_names=None,
+                             targets=None, # type: Optional[list]
+                             layer=None, # type: Optional[Union[int, str, Layer]]
+                            ):
+    # type: (...) -> Explanation
+    # FIXME: in docs rendered param order is "type, required (paramname)", should be other way around
```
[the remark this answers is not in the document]
teabolt: The problem was with the way certain docstrings looked when the docs were built. 

For example some parameter descriptions formatted like this: 
**keras.models.Model** (*estimator*)

whereas what we want is:
**estimator** (*keras.models.Model*) 

Thankfully it was an easy fix: https://stackoverflow.com/a/32700337/11555448, https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

eli5/keras/gradcam.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # output of target layer, i.e. activation maps of a convolutional layer
+    activation_output = activation_layer.output 
+
+    # differentiate ys (scalar) with respect to each of xs (python list of variables)
+    # FIXME: this might have issues in certain cases
+    # See https://github.com/jacobgil/keras-grad-cam/issues/17
+    # a fix is the following piece of code:
+    # grads = [grad if grad is not None else K.zeros_like(var) 
+    #         for (var, grad) in zip([activation_output], grads)]
```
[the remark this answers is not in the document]
teabolt: I have ran a manual test with VGG16 (the model used in https://github.com/jacobgil/keras-grad-cam/issues/17 that originally broke things) and there seems to be no issue.

https://github.com/jacobgil/keras-grad-cam/issues/17#issuecomment-423057265 and https://github.com/tensorflow/tensorflow/issues/783#issuecomment-175824168 suggest that there is something wrong with the model or the gradient calculation when `grads` contains `None` values. I can rephrase the comment, or raise an exception when this happens to let the user know?

eli5/base.py
```diff
@@ -122,6 +130,12 @@ def __init__(self,
         self.value = value
 
 
+empty_feature_weights = FeatureWeights(
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/e26997dc1a8ff82e9901d88c75288e0ddca24614

eli5/formatters/image.py
```diff
@@ -0,0 +1,266 @@
...
+            mode = 'RGB'
+        else:
+            raise ValueError('Rank 3 heatmap must have 4 channels (RGBA), '
+                             'or 3 channels (RGB). '
+                             'Got shape with {} channels'.format(channels))
+    else:
+        raise ValueError('heatmap array must have rank 2 (L, grayscale)' 
+                         'or rank 3 (RGBA, colored).'
+                         'Got: %d' % rank)
+    heatmap = (heatmap*255).astype('uint8') # -> [0, 255] int
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/7e81e527a76e96f0d2c1dac50c4a914fa74cbe8b

eli5/ipython.py
```diff
@@ -132,6 +144,27 @@ def show_prediction(estimator, doc, **kwargs):
     keyword arguments, so it is possible to get explanation and
     customize formatting in a single call.
 
+
+    (*New in x.y.z* ) # TODO
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/48378088ed025041db12efb98d4714725d6a1231

eli5/keras/explain_prediction.py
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/2296c2c1d6714b280529604457c7ea486f05c3a5

eli5/keras/gradcam.py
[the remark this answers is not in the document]
lopuhin: I see, thanks for checking!

> raise an exception when this happens to let the user know?

Sounds good. Do you mean adding an extra check somewhere between these two lines?
```
grads, = grads # grads should be a singleton list (because xs is a singleton)
grads =  K.l2_normalize(grads)
```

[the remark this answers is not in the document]
teabolt: Yeah that sounds good.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/e57c8d696c98314f4755c689b860493a274ccea6

eli5/keras.py
teabolt: @lopuhin it might be a good idea to add this straight away to not break backwards compatibility (`image` will become a "required" argument for image explanations?)

lopuhin: @teabolt I assumed we could still allow passing ``doc`` only, or do you want to make it required? Either way I don't think it's a blocker for merging this PR.

teabolt:
> > Maybe instead we could give up and declare everything scores for now, not sure.
> 
> I could make a mention in the docs that `.proba` is set if final layer is `softmax` or `sigmoid`?
> 
> Edit: though I am not sure about use cases for `.proba`

@lopuhin maybe I can add a new argument to `explain_prediction_keras`, like `is_proba`? Currently it could be confusing to the user to see `proba` set to None when their model outputs probabilities, just with different final activation.

eli5/keras.py
teabolt: I think there's a couple of advantages for making `image` required:
* We don't need code to convert `doc` to an image, which could fail and therefore needs error handling.
* If `image` is passed, we can straight away dispatch to the image explanation function. Without `image` we would have to use the rank of `doc` or something else to determine what exactly are we explaining (image or text).

However, if you think this PR is fine without the extra argument, go ahead. We can add this later.

teabolt:
> new argument to explain_prediction_keras, like is_proba

@lopuhin sorry that I keep on changing my mind regarding this scores/probabilities thing.
I thought about it and maybe the cleanest solution is to just keep everything as scores. I just don't see a real use for switching up the field in which we store the network's prediction.

lopuhin:
@teabolt no problem in changing you mind :) And I think that what you propose is quite sensible 👍

docs/source/libraries/keras.rst
```diff
@@ -0,0 +1,81 @@
...
+------------------
+
+Currently ELI5 supports :func:`eli5.explain_prediction` for Keras image classifiers.
+:func:`eli5.explain_prediction` explains image classifications through `Grad-CAM <https://arxiv.org/pdf/1610.02391.pdf>`_.
+
+The returned :class:`eli5.base.Explanation` object has some important properties:
+
+* ``image`` represents the image that is inputted into the model.
+
+* ``heatmap``  is a grayscale "localization map", roughly indicating regions of importance in the image for the predicted class.
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/20cb6ecfd376149aef91ee9c80264c38ad6a41b9

eli5/keras/explain_prediction.py
```diff
@@ -0,0 +1,225 @@
...
+    
+    # TODO: maybe do the sum / loss calculation in this function and pass it to gradcam.
+    # This would be consistent with what is done in
+    # https://github.com/ramprs/grad-cam/blob/master/misc/utils.lua
+    # and https://github.com/ramprs/grad-cam/blob/master/classification.lua
+    values = gradcam_backend(estimator, doc, targets, activation_layer)
+    weights, activations, grads, predicted_idx, predicted_val = values
+    heatmap = gradcam(weights, activations)
+
+    doc = doc[0] # rank 4 batch -> rank 3 single image
```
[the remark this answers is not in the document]
teabolt: https://github.com/TeamHG-Memex/eli5/pull/315/commits/5f0574496b53a557564c9480c3e490c39d1ace56

I believe this change is also good for checking that `batch_size=1` for this PR.

tests/utils_image.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Check that two PIL images are equal
+    pixel-by-pixel.
+    """
+    # see https://stackoverflow.com/questions/1927660/compare-two-images-the-python-linux-way
+    # compute pixel-by-pixel difference
+    diff = PIL.ImageChops.difference(im1, im2)
+    # if no difference, array is all 0's
+    diffa = np.array(diff)
+    assert np.sum(diffa) == 0
```
kmike: What do you think about changing it to 
```suggestion
    assert np.count_nonzero(diffa) == 0
```

While the current code is correct, it is not straightforward: `sum(X)` can be 0 if some elements are positive and some are negative, so this check ("array is all 0's") only works if X is non-negative, and it is far from obvious that ImageChops.difference returns abs(difference), not just a difference.

docs/source/libraries/keras.rst
```diff
@@ -0,0 +1,81 @@
...
+------------------
+
+Currently ELI5 supports :func:`eli5.explain_prediction` for Keras image classifiers.
+:func:`eli5.explain_prediction` explains image classifications through `Grad-CAM <https://arxiv.org/pdf/1610.02391.pdf>`_.
+
+The returned :class:`eli5.base.Explanation` object contains some important objects:
+
+* ``image`` represents the image that is inputted into the model.
+
+* ``heatmap``  (property of each :class:`eli5.base.TargetExplanation` in ``targets``) is a grayscale "localization map", roughly indicating regions of importance in the image for the predicted class.
```
kmike: Do we want to be more specific here, e.g. 

* what grayscale means? Is it a 2d array with numbers from 0 to 255, with floats from 0 to 1, with arbitrary floats, including negative? 
* what do these numbers represent - e.g. 
  * what do negative numbers mean for a particular target,
  * what does it mean if model predicts that the object is not of a particular class, but some numbers in heatmap are high - does it mean object is not of this class *because* of these regions, or does it mean that regions represent some "evidence" that the object belongs to a class, but evidence is not enough?

I think we may be able to explain this all here, because a specific algorithm (Grad-CAM) is mentioned above.

tests/utils_image.py
teabolt: I agree. This is a good change. I have used `np.count_nonzero` in another test to check for all 0's and it makes sense.

docs/source/libraries/keras.rst
teabolt: Sensible changes, especially about explaining what Grad-CAM actually does. I didn't want to include specific details about the ``heatmap``, etc as I thought it would change later, but it's fine for the PR I guess.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/d7aae228da838afc0fea6133a6d19c9934c1d73e

kmike: I think specific details are fine here, because these are docs for Keras models explained by Grad-CAM. You're right that .heatmap may meean something else in a different context, so probbaly we shouldn't explain it like this in some general docs about TargetExplanation data structure.

tests/test_keras_integration.py
```diff
@@ -0,0 +1,140 @@
...
+    # check show function
+    show_overlay = eli5.show_prediction(keras_clf, cat_dog_image, targets=targets)
+    assert_pixel_by_pixel_equal(overlay, show_overlay)
+
+
+@pytest.fixture(scope="function")
+def show_nodeps():
+    # set up
+    old_format_as_image = eli5.ipython.format_as_image
+    eli5.ipython.format_as_image = ImportError('mock test')
```
kmike: that's an interesting trick :) I wonder if yield should be wrapped in try/finally, to make sure teardown code is called

tests/test_keras_integration.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    original = expl.image
+    # check external properties
+    assert isinstance(overlay, Image.Image)
+    assert overlay.width == original.width
+    assert overlay.height == original.height
+    assert overlay.mode == 'RGBA'
+
+
+def assert_attention_over_area(expl, area):
```
kmike: 👍 for specific assert functions like this

eli5/sklearn/text.py
```diff
@@ -13,10 +13,11 @@
 
 
 def get_weighted_spans(doc, vec, feature_weights):
-    # type: (Any, Any, FeatureWeights) -> Optional[WeightedSpans]
+    # type: (Any, Any, Union[FeatureWeights, None]) -> Optional[WeightedSpans]
```
kmike: it seems this function requires FeatureWeights to be not None, so maybe it makes sense to keep type signature the same, but move assert to the caller code

eli5/keras/gradcam.py
```diff
@@ -0,0 +1,206 @@
...
+                         'ys: {}. xs: {}. grads: {}'.format(
+                            ys, xs, grads))
+
+    # this seems to make the heatmap less noisy
+    grads =  K.l2_normalize(grads) 
+    return grads
+
+
+def _get_target_prediction(targets, estimator):
+    # type: (Union[None, list], Model) -> K.variable
```
kmike: Optional[List]

eli5/keras/gradcam.py
```diff
@@ -0,0 +1,206 @@
...
+    # to use grads with other operations and with K.function
+    # we need to work with the actual tensors and not the python list
+    grads, = grads # grads should be a singleton list (because xs is a singleton)
+
+    # validate that the gradients were calculated successfully (no None's)
+    # https://github.com/jacobgil/keras-grad-cam/issues/17#issuecomment-423057265
+    # https://github.com/tensorflow/tensorflow/issues/783#issuecomment-175824168
+    if grads is None:
+        raise ValueError('Gradient calculation resulted in None values. '
+                         'Check that the model is differentiate and try again. '
```
kmike:
```suggestion
                         'Check that the model is differentiable and try again. '
```

eli5/keras/explain_prediction.py
```diff
@@ -0,0 +1,226 @@
...
+        :raises ValueError: if ``targets`` is a list with more than one item.
+        :raises TypeError: if ``targets`` is not list or None.
+    :type targets: list[int], optional
+
+    :param layer:
+        The activation layer in the model to perform Grad-CAM on,
+        a valid keras layer name, layer index, or an instance of a Keras layer.
+        
+        If None, a suitable layer is attempted to be retrieved. 
+        See :func:`eli5.keras._search_layer_backwards` for details.
```
kmike: I think if we're documenting a function, it makes sense to make it public - or just document the  behavior without mentioning  a function.

eli5/keras/explain_prediction.py
```diff
@@ -0,0 +1,226 @@
...
+    :param keras.models.Model estimator:
+        Instance of a Keras neural network model, 
+        whose predictions are to be explained.
+
+
+    :param numpy.ndarray doc:
+        An input image as a tensor to ``estimator``, 
+        from which prediction will be done and explained.
+
+        For example a ``numpy.ndarray``.
```
kmike: are there other supported data types, why "for example"?

docs/source/libraries/keras.rst
teabolt: Removed mention of TargetExplanation in `libraries/keras.rst`.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/359aed8c1c7852bd8388720674941a6262bf0f5b

tests/test_keras_integration.py
teabolt: https://github.com/pytest-dev/pytest/issues/2508 and https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code suggest the "addfinalizer method" to ensure tear-down executes even if:
* the test code fails
* the set up fails (currently tear down is not executed if this happens)

https://github.com/TeamHG-Memex/eli5/pull/315/commits/50f7000c2d78beaf491f1ab516e83277e2ddb5c6

eli5/sklearn/text.py
teabolt: Moved the assert to `add_weighted_spans`. Good one. Strange that https://github.com/TeamHG-Memex/eli5/search?q=get_weighted_spans did not show `add_weighted_spans`'s call to `get_weighted_spans`.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/913d415e6bb3c2b5eeedefd164ff0b43bc3e7487

eli5/keras/explain_prediction.py
teabolt: I will explicitly mention that numpy arrays are required.

I think it is possible to use other input types (https://github.com/keras-team/keras/blob/ed07472bc5fc985982db355135d37059a1f887a9/keras/engine/training.py#L1315), i.e. tensorflow tensor. However, I haven't tested with other types and I think I have some numpy dependencies in my code. Adding more input types could be a separate GitHub issue?

https://github.com/TeamHG-Memex/eli5/pull/315/commits/9d2d22af443be5b96a1d0626b78943752023cb07

eli5/keras/explain_prediction.py
teabolt: Good catch. Forgot that I made the function private.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/bcaf7ca5724189b88f0529a0ca85aa1ccf860bd9

eli5/keras/gradcam.py
teabolt: Ah I was thinking whether to use this. My thinking was that `targets` is not actually optional in the parameter list so it might be confusing. But I see that the rest of the library uses Optional (i.e. for functions that could return None) so I will change it.

https://github.com/TeamHG-Memex/eli5/pull/315/commits/e8a34f10874bf7ae563566cd6d82e69a229aa144

teabolt:
Currently the parameter `resampling_filter` (previously called `interpolation`) of `eli5.format_as_image()` takes an integer from https://pillow.readthedocs.io/en/stable/handbook/concepts.html#filters, i.e. the user passes something like `resampling_filter=PIL.Image.BOX`. It could be clearer to take the filter as a string, i.e. let the user say `resampling_filter="BOX"`?

lopuhin:
>  ``resampling_filter=PIL.Image.BOX``. It could be clearer to take the filter as a string, i.e. let the user say ``resampling_filter="BOX"``?

I think both options are fine, to me a constant looks a bit better than a string, and I think it's fine to use a PIL constant here.

lopuhin:
Thanks for a great new feature @teabolt , and thanks for review @kmike , merging 🎉

docs/source/overview.rst
```diff
@@ -45,6 +45,9 @@ following machine learning frameworks and packages:
 * :ref:`library-sklearn-crfsuite`. ELI5 allows to check weights of
   sklearn_crfsuite.CRF models.
 
+* :ref:`library-keras` - explain predictions of image classifiers
+  via Grad-CAM visualizations.
```
kmike: Sorry for a late comment: could you please copy overview.rst changes to README file in the repo root?

kmike:
Thanks @teabolt and @lopuhin, great work!

docs/source/overview.rst
teabolt: Will do that in https://github.com/TeamHG-Memex/eli5/pull/329.  Thanks

PR #324, opened 2019-07-05
Prepare 0.9.0 release

lopuhin:
I'll make 0.9.0 release once this PR is merged into master. Added a short overview of release procedure.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-0.9.2 (2019-07-05)
+0.9.0 (2019-07-05)
 ------------------
 
 * CatBoost support: show feature importances of CatBoostClassifier,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+0.9.2 (2019-07-05)
+------------------
+
+* CatBoost support: show feature importances of CatBoostClassifier,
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+0.9.2 (2019-07-05)
```
kmike: shouldn't it be 0.9.0?

docs/source/contribute.rst
```diff
@@ -17,5 +17,16 @@ from source checkout.
 
 We like high test coverage and mypy_ type annotations.
 
+Making releases
+---------------
```
kmike: What do you think about adding a note that contributors shouldn't do this themselves, that this is a memo for the maintainers? It is inconvenient to receive pull requests which bump package version, in addition to implementing a feature.

lopuhin: Good point. What do you think of the following right after the header:

```
Note: releases are made from master by eli5 maintainers.
When contributing a pull request, please do not update release notes
or package version.
```

kmike: :+1:

kmike:
Thanks for taking care of the release @lopuhin!

lopuhin:
Thank you for review @kmike ! Released https://pypi.org/project/eli5/ and tagged.

PR #325, opened 2019-07-17
Explain text predictions of Keras classifiers

teabolt:
Based on https://github.com/TeamHG-Memex/eli5/pull/315 and https://github.com/TeamHG-Memex/eli5/pull/329, we use [Grad-CAM](https://arxiv.org/pdf/1610.02391.pdf) to highlight parts of text that contribute to a prediction of a Keras classifier.

For example, we can use a call similar to this:
```
eli5.show_prediction(model, doc, 
        tokens=doc_tokens,
        pad_x=word_index['<PAD>'],
        padding='post',
        relu=False,
)
```
To roughly highlight positive and negative parts of a text (explaining the score given in sentiment analysis):
Screenshot from 2019-07-17 18-13-25
 
This PR also makes refactorings (including changes to the public API) to the code at the linked PR's.

WIP items:
* ~~Tutorial.~~
* ~~Pass CI and coverage.~~
* ~~Tests (manual and automated).~~
* ~~Docs.~~
* ~~Mypy.~~


docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+See also the tutorial for images
+(https://eli5.readthedocs.io/en/latest/tutorials/keras-image-classifiers.html).
+Certain sections such as 'removing softmax' and 'comparing different
+models' are relevant for text as well.
+
+Set up
+------
+
+First some imports
```
lopuhin: Would be nice to state which tensorflow and keras versions this is tested with, so that in the future it's possible to reproduce it.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+``tests/estimators`` package, whose source you can check for your own
+reference. You may need extra steps here to load your custom model and
+data.
+
+.. code:: ipython3
+
+    # we need to go back to top level in order to import some local ELI5 modules
+    
+    old = os.getcwd()
+    os.chdir('..')
```
lopuhin: This is for modules defined in tests, right? Then a more common way to import them would be to change the pythonpath (via ``sys.path``).

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+In this example we will have a recurrent model with word level
+tokenization, trained on the IMDB dataset
+(https://keras.io/datasets/#imdb-movie-reviews-sentiment-classification).
+The model has one output node that gives probabilities. Output close to
+1 is positive, and close to 0 is negative.
+
+See
+https://www.tensorflow.org/beta/tutorials/text/text\_classification\_rnn
+for a simple example of how to build such a model and prepare its input.
```
lopuhin: If we have a notebook or script which trains this model, would be great to link to it here.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+
+Load our test and train data. We have a module that will do
+preprocessing for us. For your own usage you may have to do your own
+preprocessing.
+
+.. code:: ipython3
+
+    import tests.estimators.keras_sentiment_classifier.keras_sentiment_classifier \
+    as keras_sentiment_classifier
```
lopuhin: would be nice to add indentation here
```suggestion
        as keras_sentiment_classifier
```

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+    ['loss', 'acc']
+    25000/25000 [==============================] - 88s 4ms/step
+
+
+
+
+.. parsed-literal::
+
+    [0.4319177031707764, 0.81504]
```
lopuhin: If there is some way to make it more clear that first value is loss and second is accuracy (or leave only accuracy), that would be nice to have.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+
+
+.. parsed-literal::
+
+    array([[0.1622659]], dtype=float32)
+
+
+
+As expected, looks pretty low accuracy.
```
lopuhin:
```suggestion
As expected, looks pretty low score.
```

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+Modify explanations with the ``relu`` and ``counterfactual`` arguments
+----------------------------------------------------------------------
+
+In the last section we only saw the "positive" words in our input, what
+made the class score "go up". To "fix" this and see the "negative" words
+too, we can pass two boolean arguments.
+
+``relu`` (default ``True``) only shows what makes the predicted score go
+up and discards the effect of counter-evidence or other classes in case
+of multiclass classification (set to ``False`` to disable).
```
lopuhin: Maybe also would be good to add how it works inside, like "it discards negative gradients" or something similar to how it's described in the paper.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
+
+
+
+
+We can see what made the network decide that is is a negative example.
+
+Another argument ``counterfactual`` (default ``False``) highlights the
+counter-evidence for a class, or what makes the score "go down" (set to
+``True`` to enable). This is mentioned in the Grad-CAM paper
+(https://arxiv.org/abs/1610.02391).
```
lopuhin: Maybe would be good to add that for binary classification, this is the same as inverting the classes.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,3685 @@
...
...sl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 99.95%); opacity: 0.80" title="0.000">&lt;PAD&gt;</span
```
lopuhin: Looks like we can strip ``pad_value`` both from the start and the end, thus removing ``padding`` argument from the API.

lopuhin: I think it would be more clear to separate them in the API, e.g. have ``pad_value`` to be a number, and ``pad_token`` to be a string.

lopuhin: If there is a notebook/script where we trained the model, would be great to link it here.

lopuhin (commented):
This looks awesome @teabolt 👍 
Just check the tutorial, I hope comments will be rendered correctly (had some trouble with github UI so had to stop with inline comments).

General comments:
- I think it would be best to default to the last convolutional layer, as for images, instead of the first - because this seems closer to how Grad-CAM is defined (else we're getting a basic gradient method), and seems to give better results (e.g instead of highlighting one letter, the whole word is highlighted). And would be good to mention this default.
- A great section which compares explanation from different layers. Maybe if a little padding is added, it would be even more clear that using a dense layer is a bad idea.
- Probably interpolation section is not that interesting compared to the rest, as there is little difference between different options and the user is unlikely to change the default, so maybe it can be ommited from the tutorial (but left in the API)

lopuhin:
OK github (or me) indeed screwed up some inline comments, let me remove and re-add them.

lopuhin:
Re-posting removed inline comments:
- regarding ``padding`` argument: looks like we can strip ``pad_value both from the start and the end, thus removing ``padding`` argument from the API.
- regarding ``pad_value`` as number or string: I think it would be more clear to separate them in the API, e.g. have ``pad_value`` to be a number (token id), and ``pad_token`` to be a string (token used for padding).
- regarding the model used for multi-class classification: if there is a notebook/script where we trained the model, would be great to link it here.

eli5/keras/explain_prediction.py
```diff
@@ -253,90 +451,253 @@ def _extract_image(doc):
...
-    doc_sh = doc.shape
-    if len(input_sh) == 4:
-        # rank 4 with (batch, ...) shape
-        # check that we have only one image (batch size 1)
-        single_batch = (1,) + input_sh[1:]
-        if doc_sh != single_batch:
-            raise ValueError('Batch size does not match (must be 1). ' 
-                             'doc must be of shape: {}, '
-                             'got: {}'.format(single_batch, doc_sh))
+def _unbatch_tokens(tokens):
```
lopuhin: FWIW I think it's fine to require tokens to be a list without any extra dimentions, but current support also looks fine to me

eli5/keras/explain_prediction.py
```diff
@@ -253,90 +451,253 @@ def _extract_image(doc):
...
     required_rank = len(model.input_shape)
     return rank == required_rank
+
+
+def _autoget_layer_text(model, character=False):
+    # type: (Model, Union[np.ndarray, list]) -> Layer
+    """Try find a suitable layer for text ``model``.
+    If ``character`` is `True`, tokenization is character-level.
+    """
+    if character:
```
lopuhin: As we discussed, let's try to return the last spatial layer in all cases instead?

lopuhin (commented):
Thanks @teabolt , left a few more comments, will continue review today

tests/estimators/keras_sentiment_classifier/keras_sentiment_classifier.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def prepare_train_test_dataset():
+    np_load_old = np.load
+    np.load = lambda *a,**k: np_load_old(*a, allow_pickle=True, **k)
+    (x_train, y_train), (x_test, y_test) = imdb.load_data(num_words=vocablen,
+                                                          seed=113,
+                                                          start_char=1,
+                                                          oov_char=2,
+                                                          index_from=3, # <-- this does not seem to work?
```
lopuhin: Probably would be nice to clarify this comment - I see that above we do ``v + 3``, is ``index_from=3`` still required, or it's a workaround? Also does code below work ok despite this line not having an effect?

tests/test_keras_integration_text.py
```diff
@@ -0,0 +1,230 @@
...
+def sentiment_input_all_pad():
+    sample = ""
+    doc, tokens = keras_sentiment_classifier.string_to_vectorized(sample, pad=True)
+    print('Input:', sample, doc, tokens, sep='\n')
+    return doc, tokens
+
+
+# model 2: ~200,000 parameters
+# For training details see
+# tests/estimators/keras_multiclass_text_classifier/keras_multiclass_text_classifier.ipynb
```
lopuhin: 👍

lopuhin (commented):
Thanks @teabolt , finished a quick scan of the PR, looks great and nice improvements all around 👍 
Will do a more thorough review tomorrow morning.

tests/estimators/keras_sentiment_classifier/keras_sentiment_classifier.py
teabolt: Just tested loading the dataset, it seems that `index_from` is the index *after* which words begin, i.e. 3 is not a valid index for a word, but 4 is. The line has no effect since 3 is the default.

Clarified the comment.

docs/source/libraries/keras.rst
```diff
@@ -10,64 +10,176 @@ and experimental neural network architectures.
...
+    - `None` for automatically taking the top prediction that the model makes.
 
-* ``doc`` is an image as a tensor that can be inputted to the model.
+* ``layer`` is the layer in the model from which the heatmap will be generated. Possible values are:
     
-    - The tensor must be an instance of ``numpy.ndarray``.
+    - An instance of ``Layer``, a layer name (str), or a layer index (int).
 
-    - Usually the tensor has the format `(batch, dims, ..., channels)` (channels last format, `dims=(height, width)`, `batch=1`, one image), i.e. `BHWC`.
+    - `None` for automatically choosing a suitable layer. Layer is searched for by going backwards from the output through the flattened list of layers. The type and the dimensions of the layer are checked. If no suitable layer can be found, the middle layer is taken.
```
lopuhin: > If no suitable layer can be found, the middle layer is taken.

Maybe it would be better to raise an error if no suitable layer was found?

eli5/keras/explain_prediction.py
```diff
@@ -167,68 +248,194 @@ def explain_prediction_keras_image(model,
...
+        With padding if ``doc`` has padding.
+
+        A Python list or a numpy array of strings. With the same length as ``doc``.
+        If ``doc`` has batch size = 1, batch dimension from tokens may be omitted.
+
+        These tokens will be highlighted for text-based explanations.
+    :type tokens: list[str], optional
+
+    :param pad_value:
+        Number for padding. If given, cuts padding off.
```
lopuhin: Probably a more common way to say this would be something like
```suggestion
        Integer identified of the padding token. If given, cuts padding off.
```

eli5/keras/explain_prediction.py
```diff
@@ -253,90 +460,203 @@ def _extract_image(doc):
...
-def _is_suitable_activation_layer(model, layer):
-    # type: (Model, Layer) -> bool
-    """
-    Check whether the layer ``layer`` matches what is required 
-    by ``model`` to do Grad-CAM on ``layer``.
-    Returns a boolean.
+# def _forward_layers(model):
+#     # type: (Model) -> Generator[Layer, None, None]
+#     """Return layers going from input to output."""
+#     return (model.get_layer(index=i) for i in range(0, len(model.layers), 1))
```
lopuhin: Can this be removed?

tests/estimators/keras_sentiment_classifier/keras_sentiment_classifier.py
lopuhin: Great, thanks for checking 👍

lopuhin (commented):
Thanks @teabolt just checked the latest changes, will check updated text tutorial next.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,4124 @@
...
+
+Now this is something. The words highlighted in green show what makes
+the score "go up", i.e. the "positive" words (check the next section to
+see how to show positive AND negative words with the ``relu`` argument).
+
+Hover over the highlighted words to see their "weight".
+
+Even though the explanation is bright green, the actual prediction is
+not very positive. Pass ``preserve_density=True`` to make the underlying
+formatter show colors "closer" to the values they represent.
```
lopuhin: I'm not sure that using ``preserve_density=True`` here is the right thing to do, from the docs:

> If preserve_density is True, then color for longer fragments will be less intensive than for shorter fragments, so that "sum" of intensities will correspond to feature weight.
> If preserve_density is None, then it's value is taken from the corresponding attribute of doc_weighted_spans.

So it makes "is" brighter because it's shorter, I'm not sure whether it's a good thing or not, but if we want to leave that in, I'd prefer to have a more accurate description of what it does here.

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,4124 @@
...
+        
+    
+    
+    
+
+
+
+
+That's unexpected. The input seems to have nothing that makes the
+predicted score *go up*.
```
lopuhin: I think it would be nice to mention here that this is explained below :)

docs/source/_notebooks/keras-text-classifiers.rst
```diff
@@ -0,0 +1,4124 @@
...
...ackground-color: hsl(120, 100.00%, 86.69%); opacity: 0.84" title="0.002">&lt;OOV&gt;</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 69.50%); opacity: 0.94" title="0.006">so</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 67.84%); opacity: 0.95" title="0.007">all</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 65.70%); opacity: 0.96" title="0.008">you</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 77.26%); opacity: 0.89" title="0.004">madison</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 60.00%); opacity: 1.00" title="0.009">fans</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 73.31%); opacity: 0.91" title="0.005">give</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 82.62%); opacity: 0.86" title="0.003">this</span><span style="opacity: 0.80"> </span><span style="background-color: hsl(120, 100.00%, 70.33%); opacity: 0.93" title="0.006">a</span><span style="opacity: 0.80"> </span><span
```
lopuhin: Maybe it would be more clear if we printed only the layers which we use for explanation.

lopuhin:
Some comments regarding the text tutorial (GH again places them on the wrong lines):

```
for layer in binary_model.layers:
    print(layer.name, layer.output_shape)
    if isinstance(layer, desired):
        ...
```

I'd suggest printing only the layers that we explain, else it's hard to understand the output below without checking the code in more detail.

lopuhin:
Regarding the text tutorial

> It looks like <PAD> had some effect on the explanation.

I think it would be great to clarify that we are removing padding prior to passing the input to the model, so most likely it's the model which gives different results with and without padding, right?

lopuhin:
Just finished the text tutorial, looks great 👍 
One more comment regarding the note at the end

> In general, this is experimental work. Unlike for images, this is not based on any paper.

I think that's a great note, what to you think of moving it to the start of the tutorial?

eli5/keras/explain_prediction.py
```diff
@@ -104,38 +152,71 @@ def explain_prediction_keras(model, # type: Model
     -------
       expl : :class:`eli5.base.Explanation`
         An :class:`eli5.base.Explanation` object for the relevant implementation.
+
+        The following attributes are supported by all concrete implementations:
+
+        * ``targets`` a list of :class:`eli5.base.TargetExplanation` objects \
+            for each target. Currently only 1 target is supported.
+        * ``layer`` the hidden activation layer that we chose to do computations with, \
+            as a string (name of the layer).
```
lopuhin: I think layer can also be an int and a Layer instance, and it's good that we support it, so maybe it would be better to remove or fix the type information here ("as a string").

eli5/keras/gradcam.py
```diff
@@ -90,58 +36,66 @@ def gradcam_backend(model, # type: Model
...
-    ``model``, ``doc``, ``targets`` parameters.
+
+    See :func:`eli5.keras.explain_prediction` for description of the
+    ``model``, ``doc``, and ``targets`` parameters.
 
     Returns
     -------
-    (weights, activations, gradients, predicted_idx, predicted_val) : (numpy.ndarray, ..., int, float)
+    (activations, gradients, predicted_idx, predicted_val) :
+        (numpy.ndarray, numpy.ndarray, int, float)
```
lopuhin: Last two arguments of returned tuple have different types in the type annotation (np.ndarrays) and in the docstring (int, float) - would be great to sync them if possible.

eli5/keras/gradcam.py
```diff
@@ -90,58 +36,66 @@ def gradcam_backend(model, # type: Model
...
+    activation_output = activation_layer.output
 
     # score for class w.r.p.t. activation layer
     grads = _calc_gradient(predicted_val, [activation_output])
 
-    # Global Average Pooling of gradients to get the weights
-    # note that axes are in range [-rank(x), rank(x)) (we start from 1, not 0)
-    # TODO: decide whether this should go in gradcam_backend() or gradcam()
-    weights = K.mean(grads, axis=(1, 2))
+    # TODO: gradcam on input layer
```
lopuhin: Could you please explain this TODO item in more detail, does this refer to some functionality that is not implemented yet, or something else?

lopuhin (commented):
A few more comments on the code

eli5/nn/text.py
```diff
@@ -0,0 +1,216 @@
...
+    """Highlight ``tokens`` in ``document``, with weights from ``heatmap``."""
+    assert len(tokens) == len(heatmap)
+    spans = []
+    running = 0  # where to start looking for token in document
+    for (token, weight) in zip(tokens, heatmap):
+        # find first occurrence of token, on or after running count
+        t_start = document.index(token, running)
+        # create span
+        t_end = t_start + len(token)
+        # why start and end is list of tuples?
```
lopuhin: It's useful to have a list here in case of bag of words approach, which is often used for linear models: in this case a feature is a word or n-gram (regardless of it's position in the document), and if this feature occurs in the document multiple times, then it will have several spans.

lopuhin (commented):
Finished checking the code, looks great, thanks @teabolt 👍

docs/source/_notebooks/keras-text-classifiers.rst
teabolt: You are right. Sorry should've checked the docs. I think  it makes sense to remove this and just mention that the user should hover over text to see the actual value.

lopuhin: Sounds good to me 👍

teabolt:
> I think it would be great to clarify that we are removing padding prior to passing the input to the model, so most likely it's the model which gives different results with and without padding, right?

Here I think it's the model that gives different results with and without padding. But in general we remove padding (using the `pad_token` or `pad_value` argument) *after* passing the input to the model.

lopuhin:
Great, thanks for clarification @teabolt 👍

eli5/keras/explain_prediction.py
teabolt: I can add support for int.

But I am not sure about Layer instance @lopuhin . I think I will use this `layer` argument with PyTorch as well, but it has a different type for a "layer" (`nn.Module`). The resulting type would be something like `Union[str, int, Any]`?

eli5/keras/gradcam.py
teabolt: Good spot. Forgot to update the docstring

eli5/keras/explain_prediction.py
lopuhin: Maybe we can say that type of "layer" argument is backend-specific?

eli5/keras/gradcam.py
teabolt: If I try to do an explanation using the first layer from mobilenet (model in the image tutorial), I get the following:
Screenshot from 2019-08-19 18-33-52

Essentially this error happens when I am using `K.function` and pass an input that is also an output ("both fed and fetched").

Regarding use cases: maybe generating adversarial images is one use?

eli5/keras/explain_prediction.py
teabolt: Sounds good to me. So do you want the type annotation to just be `Any`? In this case I will always pass the layer instance where possible instead of the name / id.

eli5/nn/gradcam.py
```diff
@@ -0,0 +1,205 @@
...
+    Returns
+    -------
+    lmap : numpy.ndarray
+        A Grad-CAM localization map,
+        with shape like the spatial portion of ``activations``.
+
+    Notes
+    -----
+    We currently make two assumptions in this implementation
+        * We are dealing with images as our input to ``model``.
```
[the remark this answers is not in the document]
teabolt: Absolutely. Good catch.

eli5/nn/text.py
teabolt: Makes sense! I will add a comment about this in `eli5.base` if you do not mind.

eli5/keras/explain_prediction.py
lopuhin: Yes, I think `Any` would be a fine annotation here, and passing they layer sounds good 👍

eli5/nn/text.py
lopuhin: That would be great, thanks!

Hellisotherpeople:
Oh my gosh I want this please merge this as soon as possible 🥰

teabolt:
@lopuhin Do you think there is anything major left to do in this PR? (besides making the CI pass)

lopuhin:
@teabolt I don't recall anything significant, and from a quick glance it looks like all review feedback is addressed so I think it's almost ready.

docs/source/libraries/keras.rst
```diff
@@ -10,64 +10,191 @@ and experimental neural network architectures.
...
+
+* ``dy/dA`` is the gradients of the output with respect to the activation maps.
+    * ``y`` may be a single target class in a classification task, i.e. a scalar value.
+* ``pool`` is the Global Average Pooling operation that takes gradients and averages them over certain axes.
+
+
+This is the formula presented in the Grad-CAM paper (https://arxiv.org/abs/1610.02391/).
+
+
+Depending on circumstances, in this library we may skip or modify some operations (like ReLU or pooling).
```
lopuhin, line 53: Could we make "Depending on circumstances" more precise here, e.g. "Depending on options described below", if that's the case? Another interpretation is that this could depend on some heuristics or something else.

lopuhin (commented):
thanks @teabolt I hope I can finish review in a week, looks really solid, left a few minor comments

.travis.yml
```diff
@@ -41,13 +41,13 @@ addons:
...
             - liblapack-dev
             - gfortran
-            - cmake
 
 before_install:
     - sudo apt-get -qq update
     - if [[ "$TOXENV" != "py35-nodeps" ]]; then sudo apt-get install graphviz; fi
 
 install:
+    - pip install --upgrade cmake
```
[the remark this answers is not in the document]
teabolt: Correct! Previously XGBoost couldn't build because CMake 3.13 is required, but apt installed 3.12.4 (https://travis-ci.org/github/TeamHG-Memex/eli5/jobs/704783435). Using pip seemed like the easiest way to install the newest version.

docs/source/libraries/keras.rst
teabolt: Added more detail. Let me know if it can be rephrased.

lopuhin: Looks great, thanks 👍

.travis.yml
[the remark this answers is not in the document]
lopuhin: Makes sense, thanks 👍

Hellisotherpeople:
Is this ever going to be merged? I'm super interested in seeing this get added to ELI5 ASAP!!!

teabolt:
> Is this ever going to be merged? I'm super interested in seeing this get added to ELI5 ASAP!!!

Hey @Hellisotherpeople . Sorry for the late reply. If you are still interested you can try installing the PR branch directly with pip:
1. If you are using virtualenv or a similar environment manager, make a clone of your environment.
2. Install the PR branch with `pip install -U git+https://github.com/teabolt/eli5.git@keras-gradcam-text`

To test that it worked run:
```
import eli5.keras
eli5.keras.explain_prediction_keras_text()
```
The output should be `TypeError: explain_prediction_keras_text() missing 2 required positional arguments: 'model' and 'doc'`

Tested this with Python 3.7 and Ubuntu 20.04.

PR #329, opened 2019-08-05
(Keras images) Add an optional image argument, and other improvements

teabolt:
API improvements for https://github.com/TeamHG-Memex/eli5/pull/315. 

For Keras image classifiers, the call for explanations is `eli5.explain_prediction(model, doc)` . We will add the ability to pass an optional `image` argument like this: `eli5.explain_prediction(model, doc, image=image)`, where `image` is a Pillow image corresponding to `doc`. In this case we will no longer use `doc` to create an image. This is useful for experimental models or unusual input shapes.

These changes should've been introduced in https://github.com/TeamHG-Memex/eli5/pull/325 but need to be merged earlier. See also https://github.com/TeamHG-Memex/eli5/pull/315#discussion_r292402958. (Originally `image` was going to be a required argument, but we decided to keep it optional).


kmike:
@teabolt I'm not sure I follow the comments in previous PR regarding this. Could you please summarize, why should image argument be required, not optional?

teabolt:
> I'm not sure I follow the comments in previous PR regarding this. Could you please summarize, why should image argument be required, not optional?

@kmike making ``image`` required makes a couple of things easier:
* We can dispatch to an image or text explanation function easier (if ``image`` is passed dispatch to image, if ``tokens`` is passed dispatch to text). Otherwise we'd need to check the shape of the ``doc`` array which may not be ideal (eg: if we later end up supporting video with array shape ``(height, width, frames)``, this has the same shape as an RGB image ``(height, width, 3)``).
* We don't need code to convert ``doc`` to an ``image``. Classic ``(height, width, channels)`` arrays can be easily converted to images, but the user would have issues if ``doc`` is something else but still represents an image.

kmike:
> Otherwise we'd need to check the shape of the doc array which may not be ideal

Interesting - in most other places we're dispatching based on model, not based on input data type. E.g. you've got ResNet => dispatch to image handling.

> Edit: I can see how my arguments are edge cases. We can try infer things from doc and if that fails the user can explicitly pass image?

Yeah, that's what I was thinking - don't require user to pass image if we can figure it out automatically; if it works in 80% cases that's fine, as soon as there is a way to pass image explicitly.

teabolt:
> you've got ResNet, with 2d convolutions and stuff => dispatch to image handling.

That is a good idea. For text we can check for Embedding, recurrent, or 1D layers. For images we'd expected 2D layers.

docs/source/libraries/keras.rst
```diff
@@ -42,6 +42,12 @@ Important arguments to :func:`eli5.explain_prediction` for ``Model`` and ``Seque
     
     - Check ``model.input_shape`` to confirm the required dimensions of the input tensor.
 
+* ``image`` Pillow image, corresponds to doc input.
+
+    - **Must be passed for image explanations.**
+            
+    - **Must have mode "RGBA".**
```
lopuhin: I think it would be more convenient for the user to accept any PIL image and then convert it to RGBA

eli5/formatters/image.py
```diff
@@ -97,8 +98,9 @@ def format_as_image(expl, # type: Explanation
         raise TypeError('Explanation image must be a PIL.Image.Image instance. '
                         'Got: {}'.format(image))
     if image.mode != 'RGBA':
-        raise ValueError('Explanation image must have mode "RGBA". '
-                         'Got image with mode: %s' % image.mode)
+        # normalize to 'RGBA'
+        warnings.warn('Converting image to RGBA.', stacklevel=2)
```
lopuhin: Why do we want to warn about this? I think usually warnings are used to warn about something that is suboptimal and likely needs to be fixed, but I think in this case it's completely fine to pass any image?

eli5/keras/explain_prediction.py
```diff
@@ -20,61 +31,44 @@
...
+                             ):
     # type: (...) -> Explanation
     """
-    Explain the prediction of a Keras image classifier.
+    Explain the prediction of a Keras classifier with the Grad-CAM technique.
 
-    We make two explicit assumptions
-        * The input is images.
-        * The model's task is classification, i.e. final output is class scores.
+    # TODO: explain Grad-CAM.
```
lopuhin: Maybe we want to either fix this TODO now, or else remove from docstring, as it would be in the docs?

eli5/keras/explain_prediction.py
```diff
@@ -85,54 +79,132 @@ def explain_prediction_keras(estimator, # type: Model
...
+        For best results, pick a layer that:
+
+        * has spatial or temporal information (conv, recurrent, pooling, embedding)
+          (not dense layers).
+        * shows high level features.
+        * has large enough dimensions for resizing over input to work.
 
-        :raises TypeError: if ``layer`` is not None, str, int, or keras.layers.Layer instance.
+
+        :raises TypeError: if ``layer`` is not None, str, int, or Kerass.layers.Layer instance.
```
lopuhin:
```suggestion
        :raises TypeError: if ``layer`` is not None, str, int, or Keras.layers.Layer instance.
```

eli5/keras/explain_prediction.py
```diff
@@ -85,54 +79,132 @@ def explain_prediction_keras(estimator, # type: Model
...
+    # Note that this function should only do dispatch
+    # and no other processing
+    if image is not None:
+        return explain_prediction_keras_image(model,
+                                              doc,
+                                              image=image,
+                                              targets=targets,
+                                              layer=layer,
+                                              )
+    elif _maybe_image(model, doc):
```
lopuhin: Not sure if it makes sense to warn in this case - I think if we say that ``image`` is optional, then we shouldn't raise warnings if user didn't pass it. If we remove the warning, then it's possible to join this with above "if" branch.

eli5/keras/explain_prediction.py
```diff
@@ -85,54 +79,132 @@ def explain_prediction_keras(estimator, # type: Model
...
+
+def explain_prediction_keras_not_supported(model, doc):
+    """
+    Can not do an explanation based on the passed arguments.
+    Did you pass either "image" or "tokens"?
+    """
+    return Explanation(
+        model.name,
+        error='model "{}" is not supported, '
+              'try passing the "image" argument.'.format(model.name),
```
lopuhin:
```suggestion
              'try passing the "image" argument if explaning an image model.'.format(model.name),
```

eli5/keras/explain_prediction.py
```diff
@@ -85,54 +79,132 @@ def explain_prediction_keras(estimator, # type: Model
...
+    for image and text explanations.
+
+
+    Returns
+    -------
+      expl : :class:`eli5.base.Explanation`
+        An :class:`eli5.base.Explanation` object for the relevant implementation.
+    """
+    # Note that this function should only do dispatch
+    # and no other processing
```
lopuhin: That's quite minor, but currently this function does more than dispatch, it also does ``image = _extract_image(doc)`` - maybe it's better to either move that to ``explain_prediction_keras_image``, or to update the comment.

eli5/keras/explain_prediction.py
```diff
@@ -143,14 +215,63 @@ def explain_prediction_keras(estimator, # type: Model
...
+    return rank == 4 or rank == 3
+
+
+def _maybe_image_model(model):
+    # type: (Model) -> bool
+    """Decide whether ``model`` is used for images."""
+    # FIXME: replace try-except with something else
+    try:
+        # search for the first occurrence of an "image" layer
+        _search_layer_backwards(model, _is_possible_image_model_layer)
```
lopuhin: Nice heuristics 👍

README.rst
```diff
@@ -51,6 +51,9 @@ It provides support for the following machine learning frameworks and packages:
 * sklearn-crfsuite_. ELI5 allows to check weights of sklearn_crfsuite.CRF
   models.
 
+* Keras_ - explain predictions of image classifiers via Grad-CAM visualizations.
```
kmike: To make it look more 😎 we may move Keras upper the list, e.g. right after scikit-learn (both here and in overview); it can be easy to miss otherwise, as with lightning / sklearn-crfsuite the list starts to be a bit obscure.

eli5/keras/explain_prediction.py
teabolt: Good catch. I will move the line to `explain_prediction_keras_image`.

eli5/keras/explain_prediction.py
teabolt: I'll have the description only in the docs to avoid repetition.

eli5/formatters/image.py
teabolt: As discussed the two warnings may help the user/developer in case of unexpected behaviour, i.e. image implementation is dispatched but wanted something else, or image is converted to RGBA but the resulting image doesn't look right). However, it's better to replace them with logging (possibly in a separate PR to add logging to the library).

tests/test_keras_integration.py
```diff
@@ -140,6 +146,13 @@ def fin():
...
\ No newline at end of file
+    doc, image = cat_dog_image
+    with pytest.warns(UserWarning) as rec:
+        expl = show_nodeps(keras_clf, doc)
+    assert 'dependencies' in str(rec[-1].message)
+    assert isinstance(expl, Explanation)
+
+
+def test_explain_prediction_not_supported():
+    res = eli5.explain_prediction(Sequential(), np.zeros((0,)))
```
kmike: is this check faling only because of doc shape, or because of model as well? If that's only doc, what do you think about adding a case where it fails because of a model?

teabolt: Exactly, here it fails because of `doc` and short-circuits. Added another test case.

teabolt:
Added a Grad-CAM image (https://github.com/teabolt/eli5/blob/keras-gradcam-img-v2/docs/source/static/gradcam-catdog.png) to the README after text highlighting image.

eli5/keras/explain_prediction.py
```diff
@@ -85,54 +76,124 @@ def explain_prediction_keras(estimator, # type: Model
...
+    :param numpy.ndarray doc:
+        Input representing an image.
+
+        Must have suitable format. Some models require tensors to be
+        rank 4 in format `(batch_size, dims, ..., channels)` (channels last)
+        or `(batch_size, channels, dims, ...)` (channels first),
+        where `dims` is usually in order `height, width`
+        and `batch_size` is 1 for a single image.
+
+        If ``image`` argument is not given, an image will be created \
```
kmike:
```suggestion
        If ``image`` argument is not given, an image will be created
```

teabolt:
Thanks @kmike. I'll have to clean up my backslashes in other PR's.

kmike:
Thanks @teabolt!

PR #334, opened 2019-08-28
Install typing only for old python

### LINKED ISSUE #326: eli5 installation crashing on python 3.7
I get `AttributeError: type object 'Callable' has no attribute '_abc_registry'` like in #273 

I am using eli5 0.9.0

It looks like this is because the `typing` dependence: https://github.com/python/typing/issues/573#issuecomment-405986724

I could send a PR if you agree to try what they propose above

GuillemGSubies:
Closes #326

Install typing only in older python versions to solve installation problems on modern python

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ setup(
         'scipy',
         'six',
         'scikit-learn >= 0.18',
-        'typing',
         'graphviz',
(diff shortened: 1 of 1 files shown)


lopuhin:
Thanks @GuillemGSubies , released as 0.10.1

PR #355, opened 2019-12-15
Upgrade version of mypy

zzz4zzz:
- Upgrade version of mypy
- Fix typing errors

### DIFF
--- a/eli5/formatters/text.py
+++ b/eli5/formatters/text.py
@@ -159,7 +159,7 @@ def _transition_features_lines(explanation):
     return [
         "",
         "Transition features:",
-        tabulate(tf.coef, headers=tf.class_names, showindex=tf.class_names,
+        tabulate(tf.coef, headers=tf.class_names, showindex=tf.class_names,  # type: ignore
                  floatfmt="0.3f"),
         ""
     ]
--- a/eli5/formatters/text.py
+++ b/eli5/formatters/text.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import
 from itertools import chain
 import six
+from tabulate import tabulate
 from typing import List, Optional, Iterator
 
 from eli5.base import Explanation, FeatureImportances
@@ -9,7 +10,8 @@ from . import fields
 from .features import FormattedFeatureName
 from .utils import (
     format_signed, format_value, format_weight, has_any_values_for_weights,
-    replace_spaces, should_highlight_spaces, tabulate)
+    replace_spaces, should_highlight_spaces)
+from .utils import tabulate as eli5_tabulate
 from .trees import tree2text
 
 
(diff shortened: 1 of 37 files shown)


eli5/keras/gradcam.py
```diff
@@ -199,7 +199,7 @@ def _validate_target(target, output_shape):
         output_nodes = output_shape[1:][0]
         if not (0 <= target < output_nodes):
             raise ValueError('Prediction target index is ' 
-                             'outside the required range [0, {}). ',
+                             'outside the required range [0, {}). '
```
lopuhin: Thanks for the catch 👍

eli5/formatters/utils.py
```diff
@@ -144,7 +144,7 @@ def tabulate(data,  # type: List[List[Any]]
 
 def format_weight(value):
     # type: (Real) -> str
-    return '{:+.3f}'.format(value)
+    return '{:+.3f}'.format(value)  # type: ignore
```
lopuhin: I see that this is due to https://github.com/python/mypy/issues/3186, maybe a better fix would be to use ``Union[int, float]`` here and below instead of the ``Real`` until that issue is fixed, especially since we use ``Real`` only in this module

eli5/formatters/text.py
```diff
@@ -159,7 +159,7 @@ def _transition_features_lines(explanation):
     return [
         "",
         "Transition features:",
-        tabulate(tf.coef, headers=tf.class_names, showindex=tf.class_names,
+        tabulate(tf.coef, headers=tf.class_names, showindex=tf.class_names,  # type: ignore
```
lopuhin: Looks like mypy is confused by local ``tabulate`` import shadowing the global import, probably a better fix would be to move ``from tabulate import tabulate`` to the top level and import ``tabulate`` from utils as ``eli5_tabulate`` or something like this?

eli5/formatters/utils.py
zzz4zzz: I changed `Real` to `float`. No need to use `Union[float, int]` ( https://www.python.org/dev/peps/pep-0484/#the-numeric-tower ).

eli5/formatters/text.py
zzz4zzz: Yes, good point. Changed.

zzz4zzz:
@lopuhin I've checked your comments and made fixes. Additionaly I've added [--ignore-missing-imports](https://mypy.readthedocs.io/en/latest/command_line.html#cmdoption-mypy-ignore-missing-imports) to mypy command line and cleaned up imports.

lopuhin:
Thanks @zzz4zzz , and ``--ignore-missing-imports`` is also a welcome addition. I'll check PR in more detail later today or tomorrow.

lopuhin:
Thanks for review @Gallaecio 👍

Issue #358, opened 2020-01-16
Permutation importance with sample weights and CV folds fails


rg2410:
Hi,

I'm using permutation importance with a sklearn Random Forest classifier where I pass sample weights and cross-validation folds. I'm getting an error in the `_cv_scores_importances` method in `permutation_importance.py` which I believe is caused because the sample weights are not sliced like the predictors and target.

Here's a code to illustrate this:

```
from sklearn.datasets import load_breast_cancer
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import KFold
from eli5.sklearn import PermutationImportance
import numpy as np

data = load_breast_cancer()
X, y = data.data, data.target
kf = KFold(n_splits=5).get_n_splits(X)
weights = np.random.rand(len(y))
model = RandomForestClassifier()

perm = PermutationImportance(model, cv=kf).fit(X, y, sample_weight=weights)
```

The error message:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-67bb1e3b5a62> in <module>
     11 model = RandomForestClassifier()
     12 
---> 13 perm = PermutationImportance(model, cv=kf).fit(X, y, sample_weight=weights)

~\AppData\Local\Continuum\anaconda3\envs\eli5_perm\lib\site-packages\eli5\sklearn\permutation_importance.py in fit(self, X, y, groups, **fit_params)
    198 
    199         if self.cv not in (None, "prefit"):
--> 200             si = self._cv_scores_importances(X, y, groups=groups, **fit_params)
    201         else:
    202             si = self._non_cv_scores_importances(X, y)

~\AppData\Local\Continuum\anaconda3\envs\eli5_perm\lib\site-packages\eli5\sklearn\permutation_importance.py in _cv_scores_importances(self, X, y, groups, **fit_params)
    214         base_scores = []  # type: List[float]
    215         for train, test in cv.split(X, y, groups):
--> 216             est = clone(self.estimator).fit(X[train], y[train], **fit_params)
    217             score_func = partial(self.scorer_, est)
    218             _base_score, _importances = self._get_score_importances(

~\AppData\Local\Continuum\anaconda3\envs\eli5_perm\lib\site-packages\sklearn\ensemble\forest.py in fit(self, X, y, sample_weight)
    328                     t, self, X, y, sample_weight, i, len(trees),
    329                     verbose=self.verbose, class_weight=self.class_weight)
[...]

PR #359, opened 2020-01-17
Sample weight sliced to work with cross validation, issue #358

rg2410:
Changes that could fix situation described in issue #358

### DIFF
--- a/eli5/sklearn/permutation_importance.py
+++ b/eli5/sklearn/permutation_importance.py
@@ -214,8 +214,12 @@ class PermutationImportance(BaseEstimator, MetaEstimatorMixin):
         cv = check_cv(self.cv, y, is_classifier(self.estimator))
         feature_importances = []  # type: List
         base_scores = []  # type: List[float]
+        weights = fit_params.get('sample_weight', None)
+        if weights is None:
+            weights = np.ones(len(y))
+        fit_params.pop('sample_weight', None)
         for train, test in cv.split(X, y, groups):
-            est = clone(self.estimator).fit(X[train], y[train], **fit_params)
+            est = clone(self.estimator).fit(X[train], y[train], sample_weight=weights[train], **fit_params)
             score_func = partial(self.scorer_, est)
             _base_score, _importances = self._get_score_importances(
                 score_func, X[test], y[test])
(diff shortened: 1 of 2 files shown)


rg2410:
Thanks for the suggestions @lopuhin, I've added them in the two last commits

eli5/sklearn/permutation_importance.py
```diff
@@ -214,8 +214,13 @@ def _cv_scores_importances(self, X, y, groups=None, **fit_params):
...
         feature_importances = []  # type: List
         base_scores = []  # type: List[float]
+        weights = fit_params.get('sample_weight', None)
+        fit_params.pop('sample_weight', None)
         for train, test in cv.split(X, y, groups):
-            est = clone(self.estimator).fit(X[train], y[train], **fit_params)
+            if weights is None:
+                est = clone(self.estimator).fit(X[train], y[train], **fit_params)
+            else:
+                est = clone(self.estimator).fit(X[train], y[train], sample_weight=weights[train], **fit_params)
```
lopuhin: This does the right thing, but it's possible to improve this a bit - instead of repeating almost the same line twice, it's possible to modify ``fit_params`` or create new ``fold_fit_params``, putting modified ``sample_weight`` there - then the same call can be done for both cases, avoiding repetition.

eli5/sklearn/permutation_importance.py
```diff
@@ -214,8 +214,13 @@ def _cv_scores_importances(self, X, y, groups=None, **fit_params):
         cv = check_cv(self.cv, y, is_classifier(self.estimator))
         feature_importances = []  # type: List
         base_scores = []  # type: List[float]
+        weights = fit_params.get('sample_weight', None)
+        fit_params.pop('sample_weight', None)
```
lopuhin: An minor improvement is possible here: ``pop`` already returns what it popped, so these two lines could be reduced to one:
```
 weights = fit_params.pop('sample_weight', None)
```

rg2410: Done

eli5/sklearn/permutation_importance.py
rg2410: have a look at the new changes and see if they make sense. Good suggestions

PR #396, opened 2020-12-09
Fixed depreciated inspect function

icfly2:
`getfullargspec` is a drop in replacement for `getargspec`, see [docs](https://docs.python.org/3/library/inspect.html#inspect.getargspec) for details


icfly2:
Surely this is due to a change in tests not related to this PR

lopuhin:
Thanks you! This were merged in eli5-org/eli5#2 and released to PyPI with v0.11

PR #407, opened 2021-06-10
Add support for latest xgboost version

aadu:
eli5 fails when booster.feature_names is `None`, which is now the default behavior in xgboost when passing in a vanilla numpy array.
```python
TypeError: 'NoneType' object is not iterable
```
See https://github.com/dmlc/xgboost/commit/9da2287ab8536e4fc35e3c6664af4597353fd257 for the relevant commit in xgboost


lopuhin:
The fix was released in 0.13

Issue #417, opened 2022-05-01
AttributeError when importing eli5 0.11.0


dvorst:
When importing eli5, I receive the error shown below. I have version 0.11.0 installed and use Python 3.10

```
Traceback (most recent call last):
  File "C:\Users\user\PycharmProjects\resume keywords\main.py", line 2, in <module>
    import eli5
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\eli5\__init__.py", line 6, in <module>
    from .formatters import (
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\eli5\formatters\__init__.py", line 9, in <module>
    from .html import format_as_html, format_html_styles
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\eli5\formatters\html.py", line 22, in <module>
    template_env = Environment(
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\jinja2\environment.py", line 363, in __init__
    self.extensions = load_extensions(self, extensions)
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\jinja2\environment.py", line 117, in load_extensions
    extension = t.cast(t.Type["Extension"], import_string(extension))
  File "C:\Users\user\AppData\Local\Programs\Python\Python310\lib\site-packages\jinja2\utils.py", line 149, in import_string
    return getattr(__import__(module, None, None, [obj]), obj)
AttributeError: module 'jinja2.ext' has no attribute 'with_'
```

PR #418, opened 2022-05-01
Added support for Jinja2 3.1.2 and scikit-learn 1.0.2

dvorst:
When using Jinja2 3.1.2 and scikit-learn 1.0.2, the following errors occurred:
- AttributeError: module 'jinja2.ext' has no attribute 'with_'
- ModuleNotFoundError: No module named 'sklearn.metrics.scorer'
- ModuleNotFoundError: No module named 'sklearn.feature_selection.base'

The changes in 'html.py', 'permutation_importance.py' and 'transform.py' are such that these errors are resolved while maintaining support for older versions of jinja2 and scikit-learn.

'test_import.py' is used to test importing eli5 for both newer and older dependency versions.

Also see: https://github.com/TeamHG-Memex/eli5/issues/417

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,0 +1,15 @@
+import jinja2
+import sklearn
+
+
+def test_import():
+	from packaging import version
+	if version.parse(jinja2.__version__) <= version.parse('3.0.3'):
+		raise ImportError('module "Jinja2" must be of a later version than "3.0.3".')
+	if version.parse(sklearn.__version__) < version.parse('1.0.2'):
+		raise ImportError('module "Jinja2" must be version "1.0.2" or later.')
+	import eli5
+
+
+if __name__ == '__main__':
+	test_import()
--- a/eli5/formatters/html.py
+++ b/eli5/formatters/html.py
@@ -18,10 +18,13 @@ from .features import FormattedFeatureName
 from .trees import tree2text
 from .text_helpers import prepare_weighted_spans, PreparedWeightedSpans
 
-
-template_env = Environment(
-    loader=PackageLoader('eli5', 'templates'),
-    extensions=['jinja2.ext.with_'])
+try:
+    template_env = Environment(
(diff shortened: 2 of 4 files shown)


tests/test_import.py
```diff
@@ -0,0 +1,15 @@
+import jinja2
+import sklearn
+
+
+def test_import():
+	from packaging import version
+	if version.parse(jinja2.__version__) <= version.parse('3.0.3'):
+		raise ImportError('module "Jinja2" must be of a later version than "3.0.3".')
+	if version.parse(sklearn.__version__) < version.parse('1.0.2'):
+		raise ImportError('module "Jinja2" must be version "1.0.2" or later.')
```
lopuhin, line 10: What is the goal of these checks? From the code above it seems that we do support older versions of these libraries? If we want to work only with versions specified here, it's better to specify this in setup.py.

lopuhin (commented):
Hi @dvorst , thanks for the fixes 👍 I didn't run the tests but changes look good to me, left one comment below. However, would you mind submitting the PR against https://github.com/eli5-org/eli5/, if you don't mind?

tests/test_import.py
dvorst: Yeah, the test file is probably not really needed, I omitted it and also slightly changed some other stuff in the PR for https://github.com/eli5-org/eli5/ (the sklearn issue for example was already resolved)

Hellisotherpeople:
Can we get a merge ASAP on this? ELI5 is now broken for most use-cases since you know, most people are running this on sklearn models... 

Also breaks my HF space: https://huggingface.co/spaces/Hellisotherpeople/Interpretable_Text_Classification_And_Clustering

lopuhin:
@Hellisotherpeople this was merged in https://github.com/eli5-org/eli5/pull/19 and released in 0.12, thanks for a reminder

